PR #234, opened 2023-10-26
feat: icon component

### LINKED ISSUE #112: Create an Icon component for use throughout the app
It's probably best to have a custom component for displaying icons to standardise them, their sizes and colours used across the app. Another consideration is that icons should be easy to re-theme.

The component should probably accept the following attributes but these can be discussed:

- name: the name of the icon, a string from a predefined list of functions, such as "next" or "home"
- color: a semantic theme colour, see [Daisy UI colours](https://daisyui.com/docs/colors/) or a literal colour value*
- alt/title/aria-label: optional description of the icon function defaulting to icon name
- size: a string from a predefined list of sizes, such as "s" or "l"

_\* Could be used for, e.g., party colours, unless there's a way to inject these into the theme as semantic colours as well…_

kaljarv:
## WHY:

Fixes #112 

The `Icon` component can be used to show icons in the app.

Additionally, a global `Color` type is defined that includes all the
DaisyUI color names available in the app.


frontend/src/lib/components/icon/icons/TipIcon.svelte
```diff
@@ -0,0 +1,37 @@
...
+
+### Usage
+
+```tsx
+  <FooIcon color="primary" aria-hidden="true"/>
+```
+-->
+
+<IconBase fill={color} {title} viewBox="0 -960 960 960" {...$$restProps}>
+  <path
```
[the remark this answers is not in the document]
kaljarv: Yeah, it's a bit problematic, because if we use a function that selects the path (but does not `import` it from an external file) then we always need to load the data for all of the icons when we use one of them. On the other hand, if we use a dynamic `import`, then there's popup.

The fact that there's some copied code doesn't seem such a huge issue for me, though, because it's very easy to use find-and-replace if there are any changes that need to be applied globally.

This said, we can go either way, of course. The popup issue can probably be relieved a bit by just importing the paths inside the `svg` element, so that it has a set size even while the `import` is still in progress.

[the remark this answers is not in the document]
kaljarv: FYI this multi-file approach was modelled after [Svelte Material icons](https://github.com/ramiroaisen/svelte-material-icons/tree/master/svelte-material-icons).

jaguarfi (commented):
I added a couple of messages, but I still think we should not add a svelte file per Icon. I understand why it is done in the https://github.com/ramiroaisen/svelte-material-icons project. But, I do not think we need to maintain the SVG files. I am okay with some lazy loading

frontend/src/lib/components/icon/icons/TipIcon.svelte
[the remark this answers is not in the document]
kaljarv: Okay, let's ditch individual components in that case and just make sure the size of the icon container stays the same during lazy loading.

As for maintaining the SVG files, I wouldn't auto-update those. We actually don't want them necessarily to be up-to-date with their original libraries (Material icons in this case for most of them), because if their style changes, it might no longer fit the design of the app or that of the custom icons.

We can get the SVGs easily from [this repo](https://github.com/marella/material-design-icons/blob/main/svg/round/arrow_circle_right.svg), for instance.

So, let's make this so that there's just one `Icon` component, which loads the necessary SVG file based on the icon name. Maybe we also ditch support for separate `fill` and `stroke` colours as well different `viewBox` sizes, because otherwise we would need to maintain a list of those info per icon separately as well.

I'll update this PR with such a version.

frontend/src/routes/_test/+page.svelte
```diff
@@ -0,0 +1,9 @@
+<script lang="ts">
```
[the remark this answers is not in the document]
kaljarv: Good catch! It resulted from the global export in the `icon/index.ts`, but that won't be a problem with the single `Icon` component anymore. Good to keep that in mind, however, when building further components.

frontend/src/lib/components/icon/icons/AddToListIcon.svelte
```diff
@@ -0,0 +1,37 @@
+<script lang="ts">
+  import {_} from 'svelte-i18n';
+  import type {IconProps} from '../Icon.type';
+  import {IconBase} from '../base';
```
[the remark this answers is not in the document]
kaljarv: That was only because the current version included the generic version of `Icon` for demonstration purposes, but that won't anyway be a problem with lazy-loading version.

frontend/src/routes/_test/+page.svelte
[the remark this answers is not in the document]
kaljarv: Edit: added a warning about such imports in the docs.

kaljarv:
@jaguarfi, @johannalaitila, I pushed a new version of the component now. Note that only the custom icons and two of the first Material icons are currently included, but I'll include the rest of the icons, if you think this approach is viable.

Also, the commit history is now a mess but I'll squash everything into three or so commits (icon component, glocal color type, design file for icons) before merging.

frontend/src/lib/components/icon/Icon.svelte
```diff
@@ -0,0 +1,108 @@
...
+    'secondary-content': 'fill-secondary-content',
+    'accent-content': 'fill-accent-content',
+    'info-content': 'fill-info-content',
+    'success-content': 'fill-success-content',
+    'warning-content': 'fill-warning-content',
+    'error-content': 'fill-error-content'
+  };
+
+  if (color != null && color in FILLS) {
+    classes += ` ${FILLS[color]}`;
```
jaguarfi, line 68: We could simply receive the colour with the following format `fill-{COLOR}` as suggested in the daisyUI docs https://daisyui.com/docs/colors/#-3.

frontend/src/lib/components/icon/Icon.svelte
```diff
@@ -0,0 +1,108 @@
...
+### Usage
+
+```tsx
+<Icon name="addToList" color="primary" 
+ aria-label="Add candidate to your list" aria-hidden="true"/>
+```
+-->
+
+<div aria-label={label} {...$$restProps}>
+  {#await import(`./svg/${folder}/${filename}.svg?raw`) then svg}
```
jaguarfi, line 97: Regarding the lazy loading, Why not add the `svg` files directly in the `/static/icons` folder? If we add the icons there, we could benefit from caching.

frontend/src/lib/components/icon/PreviewAllIcons.svelte
```diff
@@ -0,0 +1,32 @@
+<script lang="ts">
```
jaguarfi, line 1: will you remove this component when the PR is approved? or what is the use of this component?

frontend/src/lib/i18n/en.json
```diff
@@ -35,6 +35,37 @@
     "otherLanguages": "Other Languages",
     "themes": "Themes"
   },
+  "icons": {
```
jaguarfi, line 38: weren't these already defined in Strapi?

frontend/src/lib/types/global.d.ts
```diff
@@ -85,6 +85,29 @@ declare global {
     politicalExperience: string;
   }
 
+  /** These are all the DaisyUI colors supported by the application. */
+  type Color =
```
jaguarfi, line 89: Having some types for the colours may be a good idea, but it may be hard to suggest the right prefix for the colours. For example the `bg-{COLOR}`, `text-{COLOR}`and so on

jaguarfi (commented):
@kaljarv I left some comments.

frontend/src/lib/components/icon/Icon.svelte
```diff
@@ -0,0 +1,108 @@
...
+
+### Usage
+
+```tsx
+<Icon name="addToList" color="primary" 
+ aria-label="Add candidate to your list" aria-hidden="true"/>
+```
+-->
+
+<div aria-label={label} {...$$restProps}>
```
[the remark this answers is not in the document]
kaljarv: The wrapper is necessary to create an empty space of the correct size when the `svg` is loading and for applying the properties.

The other option would be to dynamically import only the contents of the `svg` element and define the `svg` tag itself in this component. It's a bit more complicated as far as I know, because we can't use Svelte's `@html` to import it because that only applies to standalone `html`. However, I made a demo of it `IconAlt`, see the `_test` page for that.

The inline point is true, I changed the class def from `inline` to `inline-block` so now it works as it should. (The `svg` itself was inline though, i.e. visible in the page DOM.)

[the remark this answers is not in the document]
kaljarv: True! I'm converting this to do so, and we'll get rid of the default translations as well.

frontend/src/lib/components/icon/Icon.svelte
```diff
@@ -0,0 +1,108 @@
+<script lang="ts">
+  import {_} from 'svelte-i18n';
+  import type {IconProps} from './Icon.type';
+  import {ICONS} from './icons';
+
+  type $$Props = IconProps;
+  export let name: $$Props['name'];
```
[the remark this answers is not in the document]
kaljarv: Can you try recreating the types? For me it works, but I had that problem at some point.


If you still don't get them, it might be related to the prop being called `name`. I could change it to `type` for example.

frontend/src/lib/components/icon/Icon.svelte
kaljarv: I think the docs refer in that case only to the logic of using the classes, not actual code. See: https://tailwindcss.com/docs/content-configuration#dynamic-class-names

The `fill-XXX` classes don't work, if you first remove them from the component (completely, not just commenting them out) and rebuild the app from scratch (I ran `yarn dev:down; yarn dev`), bc the classes might be cached.

frontend/src/lib/components/icon/Icon.svelte
kaljarv: I was checking the network usage, and it seems they are still cached even from where they currently are, if I understand correctly You can try this by jumping between the two pages in `_test` now.

frontend/src/lib/components/icon/PreviewAllIcons.svelte
kaljarv: I was going to save it so it can be used in a user guide to easily see all available components.

frontend/src/lib/i18n/en.json
kaljarv: I don't think so, but we can remove them all now anyway.

frontend/src/lib/types/global.d.ts
kaljarv: I think this type is still usable, because it can be used in component props. Those should probably anyway accept just the color without `fill` or `text` prefixed to it.

kaljarv:
I pushed a new version now which is still a bit of a mess, but please check the two different approches (`Icon` and `IconAlt`). Note, too, that the documentation is not up-to-date either.

frontend/src/lib/types/global.d.ts
jaguarfi: If we use this only for the components, could you add the types inside the folder `components`. if not maybe a comment saying that these colors are used in the components.

frontend/src/lib/components/icon/Icon.svelte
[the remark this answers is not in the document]
jaguarfi: it works now

frontend/src/lib/types/global.d.ts
kaljarv: I moved the type definition and the `FILLS` record to `$lib/components/shared/colors/`.

frontend/src/lib/components/icon/IconAlt.svelte
jaguarfi, line 1: I prefer this one over the initial `Icon`.

frontend/src/lib/components/icon/IconAlt.svelte
```diff
@@ -0,0 +1,86 @@
...
+    case 'lg':
+      classes += ' h-32 w-32';
+      break;
+    default:
+      classes += ' h-24 w-24';
+  }
+
+  if (color != null && color in FILL) {
+    classes += ` ${FILL[color]}`;
+  }
```
jaguarfi, line 53:
```suggestion
  if (color != null) {
    classes += ` fill-${color}`;
  }

```

would this work? I tested it locally, and it seems to work. This way, we do not have to manually define the `FILL` colours and we can just rely on the Color types.

frontend/src/lib/components/icon/icons.ts
jaguarfi, line 1: What about defining something like this:

```JavaScript
export const ICONS: ICONS = {
  addToList: {folder: 'material', fileName: 'playlist_add'},
  alliance: {folder: 'material', fileName: 'link'},
  candidate: {folder: 'material', fileName: 'candidate'},
  candidates: {folder: 'material', fileName: 'candidates'},
  check: {folder: 'material', fileName: 'check'},
  close: {folder: 'material', fileName: 'close'},
  collapse: {folder: 'material', fileName: 'collapse'},
  constituency: {folder: 'material', fileName: 'constituency'},
  download: {folder: 'material', fileName: 'download'},
  election: {folder: 'material', fileName: 'election'},
  expand: {folder: 'material', fileName: 'expand'},
  feedback: {folder: 'material', fileName: 'feedback'},
  filter: {folder: 'material', fileName: 'filter'},
  help: {folder: 'material', fileName: 'help'},
  home: {folder: 'material', fileName: 'home'},
  important: {folder: 'material', fileName: 'important'},
  info: {folder: 'material', fileName: 'info'},
  list: {folder: 'material', fileName: 'list'},
  menu: {folder: 'material', fileName: 'menu'},
  missingIcon: {folder: 'custom', fileName: 'missing_icon'},
  next: {folder: 'material', fileName: 'next'},
  opinion: {folder: 'custom', fileName: 'opinion'},
  party: {folder: 'material', fileName: 'party'},
  previous: {folder: 'material', fileName: 'previous'},
  results: {folder: 'material', fileName: 'results'},
  removeFromList: {folder: 'material', fileName: 'removeFromList'},
  search: {folder: 'material', fileName: 'search'},
  skip: {folder: 'custom', fileName: 'skip'},
  sort: {folder: 'material', fileName: 'sort'},
  tip: {folder: 'material', fileName: 'tip'}
};
```
If we do it this way we can avoid problems with nested folders.

jaguarfi (commented):
Please check my comments and let's try to get this ready this week :)

frontend/src/lib/components/icon/IconAlt.svelte
kaljarv: Unfortunately, it doesn't work. On your local version, did you check that the fill classes aren't still somewhere in the code even commented out and rebuilt the app (see [earlier comment](https://github.com/OpenVAA/voting-advice-application/pull/234#discussion_r1382945006))?

image

frontend/src/lib/components/icon/icons.ts
kaljarv: I'm formatting it to string tuples, so it's a bit less verbose but does the same thing.

kaljarv:
I updated this now to use the alt model.

frontend/src/lib/components/icon/IconAlt.svelte
kaljarv: There's one option for these, though, if we want to be able to use the `fill-${color}` format. We could [safelist](https://tailwindcss.com/docs/content-configuration#safelisting-classes) all of the relevant `fill`, `bg`, `text` etc. classes in `tailwind.config.cjs`. If you feel that would be a better option, we can do that as well.

frontend/src/lib/components/icon/PreviewAllIcons.svelte
kaljarv: True, I think it would be nice to have a preview directory of all of the components. I'm just wondering if it's better to keep that separate from the impromptu preview of a component within a PR, because we might run into merge issues if there are several PRs open at once and they all try to edit the `_test` route… 🤔

frontend/src/lib/components/icon/IconAlt.svelte
kaljarv: Let's change these to use the safelists.

frontend/src/lib/components/icon/Icon.type.ts
```diff
@@ -0,0 +1,24 @@
+import type {SvelteHTMLElements} from 'svelte/elements';
+import type {Color} from '$lib/components/shared/colors';
+import type {IconName} from './icons';
+
+export type IconProps = SvelteHTMLElements['div'] & {
```
jaguarfi, line 5: Should this be `SvelteHTMLElements['svg']`

frontend/src/lib/components/icon/Icon.svelte
```diff
@@ -0,0 +1,78 @@
...
+  $$restProps.class = `${classes} ${$$restProps.class ?? ''}`;
+</script>
+
+<!--
+@component
+An icon component, where the `name` property defines which icon to use.
+
+Use the other properties to set the size and color of the icon. The icon
+is `aria-hidden` by default, but that can overriden. You can also pass 
+any valid attributes to the `div` element wrapping the `svg` icon.
```
jaguarfi, line 57: there is no div wrapping the svg icon anymore

jaguarfi (approved):
Could you check the two comments I added regarding the `div` tag?

frontend/src/lib/components/icon/Icon.type.ts
kaljarv: Good catch! I changed this so many times, I forgot to change it here :D

PR #236, opened 2023-10-27
Hero emoji component

### LINKED ISSUE #235: As an Application Manager, I want to decorate pages with emoji illustrations.
Emojis are an easy way to add some illustrations to pages. Create a Svelte component to handle that.

See the Figma file for the design.

kaljarv:
## WHY:

Fixes #235

The component styles its content (an emoji) in such a way that it can be
used as an easy illustration on typical pages. The `div` produced is
`aria-hidden` by default but this can be overriden by supplying a
`"false"` value value for that attribute.

The emoji content on each page may be hard-coded or, bettes yet,
supplied using the standard way for text translations.

### Preview

Navigate to `/_test/` to see it in action.


frontend/src/lib/components/heroEmoji/HeroEmoji.svelte
```diff
@@ -0,0 +1,36 @@
+<script lang="ts">
```
jaguarfi, line 1: I think this component is too dynamic 😅. Basically, this is a `div` that accepts anything. Therefore, using this instead of a `div` doesn't provide much benefit, nor does it allow us to control what people place inside the `<HeroEmoji>` component. I know we want to provide dynamic components, but we should somehow limit what the users can do with the components. In this case, the name of the component is `heroEmoji`. Therefore, we should ensure the component only renders emojis.

https://unicode.org/emoji/charts/full-emoji-list.html

frontend/src/lib/components/heroEmoji/HeroEmoji.svelte
```diff
@@ -0,0 +1,36 @@
...
+  import type {HeroEmojiProps} from './HeroEmoji.type';
+
+  type $$Props = HeroEmojiProps;
```
jaguarfi, line 4: This is not being used

frontend/src/lib/components/heroEmoji/HeroEmoji.svelte
kaljarv: It's true that the component is quite generic, but I think there's still value to it.

In order to provide the same functionality otherwise, we would have to define a global utility class containing the necessary styling – but that is against the [recommended practice](https://tailwindcss.com/docs/reusing-styles#extracting-components-and-partials) when using Tailwind. Their recommendation is to build components instead of semantic classes. And I think the `aria-hidden` default is also a good one for this.

I don't think restricting what is rendered is actually necessary, because there might be a good reason for someone building a VAA to use some other glyph than an emoji for illustration. But if you feel strongly that we should strip out other characters, we could use a prop instead of a slot and use the [Unicode class regex](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Regular_expressions/Unicode_character_class_escape) `/\P{Emoji}/u` to remove everything else and limit the content length (although that's a bit tricky because some emoji use up multiple chars).

What do you think?

frontend/src/lib/components/heroEmoji/HeroEmoji.svelte
kaljarv: True, but it is needed to associate `HeroEmojiProps` as the component's allowed properties. The effect is not huge, though.

frontend/src/lib/components/heroEmoji/HeroEmoji.svelte
```diff
@@ -0,0 +1,36 @@
...
+Used for large emojis acting as illustrations.
+
+The content of the component is the emoji or emojis to be displayed.
+
+The content is hidden from screen readers by default. To override,
+add `aria-hidden="false"` to the tag.
+
+### Properties
+
+- Any valid attributes of a `div` element.
```
jaguarfi, line 23: it may be worth mentioning the classes needed to change the size of the emoji.

frontend/src/lib/components/heroEmoji/HeroEmoji.svelte
```diff
@@ -0,0 +1,36 @@
...
+- default: the emoji or emojis to be displayed.
+
+### Usage
+
+```tsx
+<HeroEmoji>🚀</HeroEmoji>
+```
+-->
+
+<div aria-hidden="true" {...$$restProps}><slot /></div>
```
jaguarfi, line 36: I may say this needs the `role=img` and perhaps we also need to mention the use of `aria-label` or `aria-labelledby`.

https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/img_role

jaguarfi (changes requested):
Couple of comments

frontend/src/lib/components/heroEmoji/HeroEmoji.svelte
jaguarfi: Oki, we can keep the slot and add no restrictions.

frontend/src/lib/components/heroEmoji/HeroEmoji.svelte
kaljarv: Good point! I'm adding `role="img"` to reflect that fact, but I would still keep it hidden by default because the intended use is [decorative](https://www.w3.org/WAI/tutorials/images/decorative/) only. I also added a note in the docs clarifying that.

We could ofc add a check that if `aria-hidden` is `false`, an `aria-label` is mandatory, but should we consider those kind of checks all at once or try it here already?

johannalaitila (approved):
LGTM

frontend/src/lib/components/heroEmoji/HeroEmoji.svelte
```diff
@@ -0,0 +1,41 @@
...
+
+The content of the component is the emoji or emojis to be displayed.
+
+The content is hidden from screen readers by default, because the
+intended use is for [decorative purposes](https://www.w3.org/WAI/tutorials/images/decorative/).
+To override, add `aria-hidden="false"` to the tag and also consider
+adding an `aria-label`.
+
+To change the size of the emoji, add a `text-[size]` utility class
+in the `class` property, e.g. `text-[10rem]`.
```
jaguarfi, line 24:
```suggestion
using the `class` attribute, e.g. `class="text-[10rem]"`.
```
or something like that.

jaguarfi (approved):
👍🏽

Issue #214, opened 2023-09-26
Build a reusable Page component to maintain consistent page layouts


kaljarv:
Now that we're implementing UI changes for a consistent layout (#135) – and one that's aligned with the V2.2 designs in [Figma](https://www.figma.com/file/3LDNZWOxuhhVaVPAOApfOK/Voting-Advice-Application?type=design&node-id=0%3A1&mode=design&t=IUz4yKLwLJINKMVa-1), it would be useful to have a component with slots for content that can be inserted into different parts of the page layout.

In the rough sketch below, the slots and props that a couple of different pages would use are shown.

<img width="840" alt="Page Component" src="https://github.com/OpenVAA/voting-advice-application/assets/7520743/a8933734-2c6d-48fa-9cd7-af102e187f5f">

The different slots would be:

- `headerActions`: optional action buttons inserted on the top-right
- `notification`: optional notification to show at the top of the page (see screen cap below)
- `heroImage`: optional image or other decorative content inserted before the main content
- `main`: required main content for the page. The hero image and main content expand so that the page always fills the screen vertically.
- `mainActions`: optional actions items that are placed at the bottom of the page (on mobile) for easy access

In addition, the component would accept some properties:

- `title`: text to used as the page's title using `<svelte:head><title>` and suffixed with the app's main title
- `progress`: optional number between 0–100, if progress is to be shown
- `headerClass`: optional class names to be added to the navigation
- `contentClass`: optional class names to be added to the container of the rest of the content (combining hero, main and main actions)

<img width="195" alt="image" src="https://github.com/OpenVAA/voting-advice-application/assets/7520743/1eee97b8-ab09-4ad3-92e9-493315a655a4">

Issue #60, opened 2023-01-27
Create Navigation Bar component to be used throughout the application


jaguarfi:
The Navigation Bar component is a component that can be reused throughout the application.

<img width="370" alt="image" src="https://github.com/OpenVAA/voting-advice-application/assets/7520743/213b0810-e52c-40ed-8b48-a586624b2b0a">

See [Figma Designs](https://www.figma.com/file/uY9RTeNGmkSXjBYEWMmgGn/Election-Compass?node-id=297%3A5805&t=ZM8SK3lSAyXPV506-0)

kaljarv:
Closed by #258

PR #238, opened 2023-11-01
feat: navigation component

kaljarv:
## WHY:

Necessary for #214 and related to #60.

The `Navigation` component takes as its input a list of `NavigationItems` and outputs a properly formatted navigation element with the items as links or buttons depending on whether a `href` or `onclick` property is provided for them.

The logic for creating the actual items is separated from the component so that the lists of `NavigationItems` can be contained in stores placed in the `$voter` and `$candidate` libraries.

The component will mainly be used in the page templates.

### DIFF
--- a/frontend/src/lib/components/navigation/NavGroup.svelte
+++ b/frontend/src/lib/components/navigation/NavGroup.svelte
@@ -1,0 +1,43 @@
+<script lang="ts">
+  import type {NavGroupProps} from './NavGroup.type';
+
+  type $$Props = NavGroupProps;
+
+  // Merge classes into $$restProps
+  $$restProps.class = `before:content-[''] before:mx-16 before:my-md before:block before:border-t-md before:border-t-[var(--line-color)] ${
+    $$restProps.class ?? ''
+  }`;
+</script>
+
+<!--
+@component
+TBA
+
+### Slots
+
+- default: The contents of the navigation group. Should be mostly 
+  `<NavItem>` components.
+
+### Properties
+
+- Any valid attributes of a `<section>` element.
+
+### Usage
+
+```tsx
+<NavGroup>
+  <NavItem href="/info" icon="info">Show info</NavItem>
+  <NavItem on:click={(e) => foo(e)}>Do foo</NavItem>
+  <div>Some other content</div>
+</NavGroup>
+```
+-->
+
+<!-- We use a <section> with an Aria role instead of a <ul> or similar
+  because otherwise generic content passed via the slot might include
+  invalid content for such an element. See: 
+  https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/listitem_role -->
+
+<section role="list" {...$$restProps}>
+  <slot />
+</section>
--- a/frontend/src/lib/components/navigation/NavItem.svelte
+++ b/frontend/src/lib/components/navigation/NavItem.svelte
@@ -1,0 +1,62 @@
+<script lang="ts">
+  import {Icon} from '$lib/components/icon';
+  import type {NavItemProps} from './NavItem.type';
+
+  type $$Props = NavItemProps;
+
+  export let icon: $$Props['icon'] = undefined;
+  export let text: $$Props['text'];
+
+  // Element type
+  const tagName = $$restProps['href'] ? 'a' : 'button';
+
+  // Merge classes into $$restProps
+  let classes =
+    'flex items-center gap-md px-16 py-md min-h-touch min-w-touch !text-neutral hover:bg-base-200 active:bg-base-200';
+  if (!icon) {
+    // This corresponds to the width of an icon (24/16 rem) and the gap
+    // between the icon and the text (md = 10/16 rem)
+    classes += ' pl-[3.125rem]';
+  }
+  $$restProps.class = `${classes} ${$$restProps.class ?? ''}`;
+</script>
+
+<!--
+@component
+Outputs a navigation item for use inside a `<NavGroup>` which in turn is
+used within a `<Navigation>` component.
+
+The item is rendered as an `<a>` element if `href` is supplied. Otherwise
+a `<button>` element will be used. Be sure to provde an `on:click` event 
+handler or other way of making the item interactive.
+
+### Properties
+
+- `href`: The URL to navigate to. If this is not supplied be sure to
+  provde an `on:click` event handler or other way of making the item
+  interactive.
+- `icon`: An optional `IconName` of the icon to use. @default `undefined`
+- `text`: A required text to display.
+- Any valid attributes of either an `<a>` or `<button>` element depending
+  whether `href` was defined or not, respectively.
+
+### Usage
+
+```tsx
+<NavItem href="/info" icon="info" text="Show info"/>
+<NavItem on:click={(e) => foo(e)} text="Do foo"/>
+```
+-->
+
+<!-- We use a div with an Aria role instead of a `<li>` because we don't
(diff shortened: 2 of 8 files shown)


kaljarv:
In order for the component to be reusable for both the Voter and the Candidate App, the logic of producing the actual navigation items is now separated from the layout.

Note that there are two `tmp:` commits that will be ditched before merging.

frontend/src/lib/components/navigation/Navigation.svelte
```diff
@@ -0,0 +1,72 @@
...
+```tsx
+<script lang="ts">
+  import {navigationItems} from '$voter/navigation';
+</script>
+
+<Navigation items={$navigationItems} aria-label="Main navigation" />
+```
+-->
+
+<nav {...$$restProps}>
```
jaguarfi, line 32: I may have a big suggestion 😅: I think, we should have two components, the `Navigation` component, and a `NavItem` component. Something like this:

```Svelte
<script>
	import { Navbar, NavbarItem } from 'yesvelte'
</script>

<Navivation>
	<NavItem title="Home" />
	<NavItem title="Features" />
	<NavItem title="Pricing" />
	<NavItem title="Disabled" disabled />
</Navigation>
```

This way we can assign classes and the `aria` properties to each navigation Item. With the current implementation, it is not possible to pass properties to the navigation Items. 

Some inspiration:
https://mui.com/material-ui/react-menu/
https://www.yesvelte.com/docs/navbar

jaguarfi (changes requested):
I just added a couple of comments. But I strongly believe we have to add a NavItem component. I could push the branch where I am refactoring the tabs so that you can see what I am trying to do. But, in the end, the new Tab component will look like this.
image

With this approach, we can manage the aria properties easily, add styles to each subcomponent and make the component more dynamic.

Please let me know what you think about this.

frontend/src/lib/components/navigation/Navigation.svelte
jaguarfi: https://joyofcode.xyz/svelte-context-with-stores

kaljarv: I'll convert it to multiple components! I think I'll also need a `NavGroup` which is basically a glorified `ul` :smile

kaljarv:
Okay, this version now has multiple subcomponents for building menus. I also added one for the Voter App.

@jaguarfi, is this something in the way you were thinking of?

frontend/src/lib/components/navigation/NavItem.svelte
```diff
@@ -0,0 +1,75 @@
...
+
+```tsx
+<NavItem href="/info" icon="info">Show info</NavItem>
+<NavItem onclick={(e) => foo(e)}>Do foo</NavItem>
+```
+-->
+
+<!-- We use a div with an Aria role instead of a <li> because we don't
+  place the items within a valid parent for a <li> -->
+<div role="listitem">
```
jaguarfi, line 59: shouldn't the role be `menuitem`

frontend/src/lib/voter/components/navigation/VoterNav.svelte
```diff
@@ -0,0 +1,45 @@
+<script lang="ts">
```
jaguarfi, line 1: We don't need this component. As the navigation for the voter app or the candidate app is defined only once, we can define it on the actual route page. Also, if we need to render some nav items dynamically based on the application state it will make more sense to define the Navigation menu in the pages.

jaguarfi (changes requested):
A couple of comments regarding the roles for the components. We may need to also allow the navigation with the arrow keys.

https://www.w3.org/WAI/ARIA/apg/patterns/menubar/examples/menubar-navigation/

frontend/src/lib/components/navigation/NavItem.svelte
```diff
@@ -0,0 +1,75 @@
+<script lang="ts">
+  import {error} from '@sveltejs/kit';
+  import {Icon} from '$lib/components/icon';
+  import type {NavItemProps} from './NavItem.type';
+
+  type $$Props = NavItemProps;
+
+  export let icon: $$Props['icon'] = undefined;
+  export let href: $$Props['href'] = undefined;
+  export let onclick: $$Props['onclick'] = undefined;
```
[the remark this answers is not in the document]
kaljarv: Okay, we can do that. The only issue is that we cannot confirm whether button will actually do anything if we rely on `on:click` or a similar handler but I guess we just have to trust the developers in that 😄

frontend/src/lib/components/navigation/NavGroup.svelte
```diff
@@ -0,0 +1,43 @@
...
+</NavGroup>
+```
+-->
+
+<!-- We use a <section> with an Aria role instead of a <ul> or similar
+  because otherwise generic content passed via the slot might include
+  invalid content for such an element. See: 
+  https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/listitem_role -->
+
+<section role="list" {...$$restProps}>
```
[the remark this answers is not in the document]
kaljarv: The ARIA guidelines say that a `menu` should contain more complex logic (think of the menu in an application) than just navigation.

frontend/src/lib/components/navigation/NavItem.svelte
kaljarv: See above.

frontend/src/lib/voter/components/navigation/VoterNav.svelte
kaljarv: In my opinion, it depends a bit on how the page templates that will be added later are structured, but we can discuss this related to that PR once I get it started. Even there, though, I think it might make sense that navigation was its separate file, i.e., component, because it's logically a bit different from the page template itself and would make for nicer separation in terms of possible conflicts with changes to files.

Anyway, I can leave it out from this PR.

kaljarv:
> We may need to also allow the navigation with the arrow keys.

I'm not sure if they will be very helpful here because the navigation menu is one-dimensional. Note that the linked example explicitly states that it is more complex than a typical navigation menu would be.

I'd suggest we make the arrow navigation optional and later implement that with a generic Svelte `use` directive, which can be applied in other similar situations as well.

kaljarv:
Update: I separated the `a`/ `button` logic from `NavItem` and moved it into its own component, bc that will be used also for button components, which are unrelated to navigation.

Also, I noticed that the `HTMLAttributes<T>` prop definitions do not actually contain all the possible properties and converted those to the correct `SvelteHTMLElements['nav']` format. This change will need to be made to all other typings using that format as well.

frontend/src/lib/components/navigation/NavItem.svelte
```diff
@@ -0,0 +1,59 @@
+<script lang="ts">
+  import {error} from '@sveltejs/kit';
```
johannalaitila, line 2: Is this required anymore? See the Github warning

kaljarv: True! I saw the warning as well and forgot to update this :D It's now corrected. I also improved the documentation a tiny bit.

kaljarv:
Based on discussion with @jaguarfi, I'll remove the `ActionItem` component and merge the logic into `NavItem`. The same kind of logic will be built into the buttons PR in a base component for them so that these two PRs don't depend on each other.

jaguarfi (approved):
Please remove the commits that are not part of this PR.

Sidenote: The keyboard navigation may need to be improved but it should be fine for now.

PR #247, opened 2023-11-09
Use docker volume over bind mounts for postgres

TrixterTheTux:
## WHY:

Currently, some people have issues with starting the project due to the Postgres container's data folder having changed ownership due to mismatching users between the host and the container. This leads to people having to run "chmod -R 777" each time when testing some of the changes.

The bind mount has been replaced with volumes so that Docker will be fully responsible for figuring out the folder's correct permissions. Additionally, I've added `--force-recreate` flag for the dev commands so that `.env` and docker compose changes always propagate.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
     "scripts": {
-        "dev":"docker compose -f docker-compose.dev.yml up -d --build",
-        "dev:attach":"docker compose -f docker-compose.dev.yml up --build",
+        "dev":"docker compose -f docker-compose.dev.yml up -d --force-recreate --build",
+        "dev:attach":"docker compose -f docker-compose.dev.yml up --force-recreate --build",
         "dev:down": "docker compose -f docker-compose.dev.yml down -v --remove-orphans",
         "dev:stop": "docker compose -f docker-compose.dev.yml stop",
         "prepare":"husky install",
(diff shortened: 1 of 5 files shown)


samuliys (approved):

package.json
```diff
@@ -1,7 +1,7 @@
 {
     "scripts": {
-        "dev":"docker compose -f docker-compose.dev.yml up -d --build",
-        "dev:attach":"docker compose -f docker-compose.dev.yml up --build",
+        "dev":"docker compose -f docker-compose.dev.yml up -d --force-recreate --build",
```
jaguarfi, line 3: I'm not sure about forcing container recreation. Of course, if we recreate the container we can make sure that we are developing on a clean slate, but it will also take more time to initialise all the containers every time we need to start our development. @johannalaitila do you have an opinion on this?

johannalaitila: It's fine for me because I think you would not run npm run dev everytime you'd wanna start the application, instead just stop/start the container directly from Docker after generating it for the first time. At least that's what I'm used to...

johannalaitila:
Looks good to me, but you'd wanna probably push these changes already to main and then just rebase candidate-app off of this, since others might have the same issues with other parts of development. Could you change the PR target branch?

TrixterTheTux:
> Looks good to me, but you'd wanna probably push these changes already to main and then just ~rebase candidate-app off of this~ merge changes from main to candidate-app, since others might have the same issues with other parts of development. Could you change the PR target branch?

This should be done.

johannalaitila (approved):
LGTM

PR #251, opened 2023-11-13
feat: button components

### LINKED ISSUE #250: Create reusable Button components for use throughout the application
Buttons are an important part of the application and they should look and behave in a similar manner throughout it. There are several different types of buttons ranging form the main button at the bottom of the page to those with just an icon in the header.

DaisyUI's `btn` classes serve as a good starting point for buttons, but they demand further design tweaks, which are best accomplished by creating Svelte components for them.

### Main button
image

### Vertical button with icon
image

### Icon button
image

### Horizontal button with icon
image

kaljarv:
## WHY:

Fixes #250 

Introduces three components for common button types used throughout the application as well as a base component for these:

- `ButtonBase`: the base component for these buttons
- `MainButton` for the main action button of a page
- `SecondaryButton` for secondary actions, such as those under an opinion questions answers or for sorting a list
- `IconButton` for buttons that only contain an icon

To see the buttons in action, see the temporary `_test` route.

### DIFF
--- a/frontend/src/lib/components/button/Button.svelte
+++ b/frontend/src/lib/components/button/Button.svelte
@@ -1,0 +1,133 @@
+<script lang="ts">
+  import {Icon} from '$lib/components/icon';
+  import type {ButtonProps} from './Button.type';
+
+  type $$Props = ButtonProps;
+
+  export let text: $$Props['text'];
+  export let variant: $$Props['variant'] = 'normal';
+  export let icon: $$Props['icon'] = null;
+  export let iconPos: $$Props['iconPos'] = 'right';
+  export let color: $$Props['color'] = 'primary';
+
+  // Check iconPos
+  if (variant === 'main' && (iconPos === 'top' || iconPos === 'bottom')) {
+    iconPos = 'right';
+  }
+
+  // Set default title and aria-label for icon variant
+  if (variant === 'icon') {
+    $$restProps['aria-label'] ??= text;
+    $$restProps['title'] ??= text;
+  }
+
+  // Element type
+  const tagName = $$restProps['href'] ? 'a' : 'button';
+
+  // Build classes
+  // 1. Base classes
+  let classes =
+    'btn min-h-touch min-w-touch h-auto flex items-center justify-center gap-y-6 gap-x-4';
+
+  // 2. Variant-defined classes
+  classes += variant === 'main' ? ' w-full max-w-md' : ' btn-ghost';
+
+  // 3. Icon position
+  switch (iconPos) {
+    case 'top':
+      classes += ' flex-col';
+      break;
+    case 'bottom':
+      classes += ' flex-col-reverse';
+      break;
+    case 'left':
+      classes += ' flex-row';
+      break;
+    case 'right':
+      classes += ' flex-row-reverse';
+      break;
+  }
+
+  // 4. Set color
+  if (color) {
+    // For the main button type we can use the btn-primary etc. classes,
+    // for the others, we just set the text color
+    classes += ` ${variant === 'main' ? 'btn' : 'text'}-${color}`;
+  }
+
+  // 5. Merge classes into $$restProps
+  $$restProps.class = `${classes} ${$$restProps.class ?? ''}`;
+
+  // 6. Finally, define the class for the text label
+  let labelClass: string;
+  if (variant === 'main') {
+    labelClass = 'flex-grow text-center';
+    if (icon) {
+      // If an icon is used, add left or right margin so that the text is
+      // nicely centered: ml/r is calculated so that it is the sum of the
+      // gap (4) and icon widths (24) = 28/16 rem
+      labelClass += iconPos === 'right' ? ' ml-[1.75rem]' : ' mr-[1.75rem]';
+    }
+  } else if (icon && (iconPos === 'top' || iconPos === 'bottom')) {
+    // We use the small-label class only in vertical buttons with an icon.
+    // The color needs to be separately applied here, bc small-label sets
+    // the text color to secondary
+    labelClass = `small-label text-${color}`;
+  }
+</script>
+
+<!--
+@component
+A component for buttons that mostly contain text and an icon. Use the
+`variant` prop to specify the button type. When using an `icon`, use
+`iconPos` to set the position of the icon relative to the text.
+
+- `main`: A large, prominent button that is used for the main action of
+  the page. In general, there should only be one of these on a page.
+- `icon`: A button containing only an icon. Note that you still need to
+  provide the `text` property, which will be used as the `aria-label` 
+  and `title` of the button.
+- `normal`: The default button type, which usually consists of an icon
+  and text. The styling for these uses DaisyUI's `btn-ghost` class, i.e.
+  the button does not have a background color.
+
+The button is rendered as an `<a>` element if `href` is supplied. 
+Otherwise a `<button>` element will be used. Be sure to provide an 
+`on:click` event handler or other way of making the item interactive.
+
+### Properties
+
+- `href`: The URL to navigate to. If this is not supplied be sure to
+  provide an `on:click` event handler or other way of making the item
+  interactive.
+- `text`: The required text of the button. If `variant` is `icon`, the
+  text will be used as the `aria-label` and `title` for the button. You
+  can override both by providing them as attributes, e.g. 
+  `aria-label="Another text"`.
+- `variant`: The type of the button.
+- `icon`: The name of the icon to display.
+- `color`: The color of the icon.
+- `iconPos`: The position of the icon relative to the text.
+- Any valid attributes of either an `<a>` or `<button>` element depending
+  whether `href` was defined or not, respectively.
+
+### Usage
+
+```tsx
+  <Button on:click={next} variant="main" icon="next"
+   text="Continue"/>
+  <Button on:click={skip} icon="skip" iconPos="top" color="secondary"
+   text="Skip this question"/>
+  <Button on:click={addToList} variant="icon" icon="addToList" 
+   text="Add to list"/>
+```
+-->
+
+<svelte:element this={tagName} on:click {...$$restProps}>
+  {#if icon}
+    <Icon name={icon} />
+  {/if}
+  {#if variant !== 'icon'}
+    <div class={labelClass}>{text}</div>
+  {/if}
+</svelte:element>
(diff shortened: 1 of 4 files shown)


johannalaitila (approved):
LGTM

frontend/src/lib/components/button/base/ButtonBase.svelte
```diff
@@ -0,0 +1,49 @@
...
+handler or other way of making the item interactive.
+
+### Slots
+
+- default: The content of the button.
+
+### Properties
+
+- `href`: The URL to navigate to. If this is not supplied be sure to
+  provde an `on:click` event handler or other way of making the item
```
jaguarfi, line 29: provide

frontend/src/lib/components/button/base/ButtonBase.svelte
```diff
@@ -0,0 +1,49 @@
...
+  interactive.
+- Any valid attributes of either an `<a>` or `<button>` element depending
+  whether `href` was defined or not, respectively.
+
+### Usage
+
+```tsx
+<ButtonBase href="/info" icon="info">
+  <Icon name="info"/> 
+  Show info
```
jaguarfi, line 39: is this accurate? do we need to pass the `<Icon>`  and the `icon` attribute?

frontend/src/lib/components/button/iconButton/IconButton.svelte
```diff
@@ -0,0 +1,44 @@
+<script lang="ts">
+  import {Icon} from '$lib/components/icon';
+  import {ButtonBase} from '../base';
+  import type {IconButtonProps} from './IconButton.type';
+
+  type $$Props = IconButtonProps;
+
+  export let icon: $$Props['icon'];
+  export let title: $$Props['title'];
```
jaguarfi, line 9: Should we remove this title property, as the title will be passed in the `restProps`?

frontend/src/lib/components/button/iconButton/IconButton.svelte
```diff
@@ -0,0 +1,44 @@
...
+- Any valid properties of an `<ButtonBase>`.
+
+### Usage
+
+```tsx
+  <IconButton on:click={showHelp} icon="help" title="Show help"/>
+```
+-->
+
+<ButtonBase aria-label={title} {title} {...$$restProps} on:click>
```
jaguarfi, line 42: I think we do not need to make the `aria-label` equal to the title. I think we can remove the `aria-label` and let the developer decide whether or not they want to pass the aria label.

frontend/src/lib/components/button/iconButton/IconButton.svelte
jaguarfi, line 1: Do we need this Icon button? Could we just use the SecondaryButton and make the button label (`text`) optional?

frontend/src/lib/components/button/secondaryButton/SecondaryButton.svelte
```diff
@@ -0,0 +1,80 @@
...
+  }
+
+  // Merge classes into $$restProps
+  $$restProps.class = `${classes} ${$$restProps.class ?? ''}`;
+</script>
+
+<!--
+@component
+Use for secondary action buttons that usually have an icon and a small 
+labe. These are used, for example, below the answering options for a
```
jaguarfi, line 47: small label

frontend/src/lib/components/button/secondaryButton/SecondaryButton.svelte
jaguarfi, line 1: I don't know if secondaryButton is a good name 😅. I may say that MainButton can be just Button and the SecondaryButton can be something else. But, it is up to you!

frontend/src/lib/components/button/secondaryButton/SecondaryButton.svelte
```diff
@@ -0,0 +1,80 @@
...
+labe. These are used, for example, below the answering options for a 
+question and the sorting options for lists (with `iconPos="left"`).
+
+In contrast to the `<MainButton>` component, there can be multiple 
+secondary buttons on a page.
+
+### Properties
+
+- `href`: The URL to navigate to. If this is not supplied be sure to
+  provde an `on:click` event handler or other way of making the item
```
jaguarfi, line 56: provide

jaguarfi (changes requested):
check my comments!

frontend/src/lib/components/button/secondaryButton/SecondaryButton.svelte
```diff
@@ -0,0 +1,80 @@
...
+  import {Icon} from '$lib/components/icon';
+  import {ButtonBase} from '../base';
+  import type {SecondaryButtonProps} from './SecondaryButton.type';
+
+  type $$Props = SecondaryButtonProps;
+
+  export let icon: $$Props['icon'];
+  export let text: $$Props['text'];
+  export let iconPos: $$Props['iconPos'] = 'left';
+  export let color: $$Props['color'] =
```
jaguarfi, line 11: why does the color depend on the position?

frontend/src/lib/components/button/base/ButtonBase.svelte
kaljarv: No, that was a typo 😄

frontend/src/lib/components/button/secondaryButton/SecondaryButton.svelte
kaljarv: It's certainly not the perfect name 😄 I would keep `MainButton` because it semantically it kind of implies that there should only be one of those on a page, which is explained in the docs. Maybe `SecondaryButton` could then be just a `Button`? I was avoiding that because of the possible mixup with`<button>` but maybe that's unnecessary.

frontend/src/lib/components/button/secondaryButton/SecondaryButton.svelte
kaljarv: Good point! It was related to the different default styling between the button types, but it's not very logical, so I'll change color to always default to `'primary'`.

frontend/src/lib/components/button/iconButton/IconButton.svelte
kaljarv: See comment below.

frontend/src/lib/components/button/iconButton/IconButton.svelte
kaljarv: The reason here is that `title` is [not an accessible description](https://www.w3.org/WAI/tutorials/forms/labels/#using-the-title-attribute) for the function of the button, but it does act as a tooltip in many browsers. And because the `IconButton` doesn't contain any text content, we need some label for it.

frontend/src/lib/components/button/iconButton/IconButton.svelte
kaljarv: The reason why `IconButton` is its own component, is the need for an accessible label (see comment above), but I agree it has a lot of similarities to `SecondaryButton`. If we combined the two, I think the easiest option would be to:

1. Still keep `text` required. In an icon only button, this would become the `title` and `aria-label`.
2. Change `iconPos?: 'top' | 'bottom' | 'left' | 'right';` to smth like `type?: 'iconOnly' | 'iconTop' | 'iconBottom' | 'iconLeft' | 'iconRight' | 'textOnly';`

What do you think?

kaljarv:
I now added a version combining all of the buttons into one component, which is currently a bit confusingly in `$lib/components/button/(Button.svelte)`. The `_test` route now showcases both the new one and the old components.

Please check which one you prefer and I'll update the PR based on that. Note that the documentation for the unitary component is not up to snuff yet.

frontend/src/lib/components/buttonAlt/Button.svelte
```diff
@@ -0,0 +1,127 @@
...
+  the page. In general, there should only be one of these on a page.
+- `icon`: A button containing only an icon. Note that you still need to
+  provide the `text` property, which will be used as the Aria label for
+  the button.
+- `normal`: The default button type, which usually consists of an icon
+  and text. The styling for these uses DaisyUI's `btn-ghost` class, i.e.
+  the button does not have a background color.
+
+The button is rendered as an `<a>` element if `href` is supplied. 
+Otherwise a `<button>` element will be used. Be sure to provde an
```
jaguarfi, line 90: be sure to **provide** ...

frontend/src/lib/components/buttonAlt/Button.svelte
```diff
@@ -0,0 +1,127 @@
...
+
+- `main`: A large, prominent button that is used for the main action of
+  the page. In general, there should only be one of these on a page.
+- `icon`: A button containing only an icon. Note that you still need to
+  provide the `text` property, which will be used as the Aria label for
+  the button.
+- `normal`: The default button type, which usually consists of an icon
+  and text. The styling for these uses DaisyUI's `btn-ghost` class, i.e.
+  the button does not have a background color.
+
```
jaguarfi, line 88: can we mention that people can change the color of the text using the  `text-{COLOR}` utility classes?

jaguarfi (changes requested):
I prefer the buttonAlt proposal, you can get rid of the other ones.

frontend/src/lib/components/button/iconButton/IconButton.svelte
jaguarfi: Yes, but the `title` and the `aria-label` are different things and have different purposes. One is targeted at visual users, and the other one is targeted at assistive technology. Therefore, they cannot be the same. Sometimes, you just need to target assistive technology without affecting the visuals.

kaljarv: I agree on a general level, but in this very specific case – a button with an icon with no text – I think it's very hard to come up with any real-world situation where anybody would want those two properties to be different 🤔. Both state the function of the button in text to those who cannot understand the icon (`title`) or see it (`aria-label`). Moreover, to my understanding there's no situation when both would be available to the same user.

Anyway, I'm changing the behaviour in the unitary component thus so both can be overriden, but we still provide useful defaults in most cases:

```ts
if (variant === 'icon') {
  $$restProps['aria-label'] ??= text;
  $$restProps['title'] ??= text;
}
```

jaguarfi: Maybe the default for the icon could be the actual icon name as most of them are quite descriptive.

jaguarfi (approved):
I added just one comment regarding the default for the  `aria-label` when the button is icon only. Otherwise it looks good,

frontend/src/lib/components/button/iconButton/IconButton.svelte
kaljarv: That's a bit problematic, because the names are not localized. What we could do would be to use a localized default name for each icon but those we don't have because we didn't want to bind the `Icon` component to the backend…

In this case, I would just use `text`, which is a required property. I think it's useful to require that whoever uses the component explicitly adds a textual description of the function for accessibility's sake.

PR #258, opened 2023-11-16
feat: page templates

### LINKED ISSUE #214: Build a reusable Page component to maintain consistent page layouts

kaljarv:
## WHY:

Fixes #214 

Introduces a base template for building page layouts (`Page`) as well as three subtemplates that can be used in different parts of the application. Additionally, these templates conform to the latest UI designs in Figma.

To see these in action, go to `_test` and click on the links demoing the different templates. NB! The existing routes have not yet been converted to use the templates.


kaljarv:
This PR is still a bit in progress, but comments would be helpful already now.

johannalaitila:
Could you add something like this as a fallback to root layout:
`<svelte:head>
  <title>{$page.data.appLabels.appTitle}</title>
</svelte:head>`

Otherwise there's a small delay when titles are lazy loaded or if title is missing and just the URL is shown in the title which is not ideal.

kaljarv:
I updated this now so that more of the general logic is handled by the base template `Page`. The biggest change is that there no longer is a `nav` slot for inserting the navigation contents. These are instead output directly by the `Page` template itself. In order for this to work, a couple of things needed to be changed:

1. Because the templates are common to both the Voter and Candidate apps, the template now queries a new store `appType`, which tells the app to use.
2. This information is used to switch between two different navigation menus.
3. The menus are separated into their own "components" mainly to move that logic out from the template file so that it's easier to update the parts separately. The navigation components are not supposed to be used elsewhere and are not exported in the module's global exports.

The logic for `appType` is perhaps not the most elegant. It's contained in a `paths.ts` file, which can also contain the names of the paths used within the app. I think that having such a global source of truth for all of the paths might well be useful in the future instead of hard-coding them.

Another option for `appType` spefically is, however, that it be a `Writable` store into which each App's outermost layout writes the currently used app. I.e., when we navigate to `/candidate`, a `/candidate/layout.svelte` file could call the `$appType = 'candidate'`.

Any opinions on this?

kaljarv:
Update: I added a new folder `parts` under templates that contains template parts, which are components that depend on the application.

frontend/src/lib/templates/frontPage/FrontPage.svelte
jaguarfi, line 1: Something off with the skip Navigation
skip

It changes the look of the page.

frontend/src/lib/templates/frontPage/FrontPage.svelte
```diff
@@ -0,0 +1,113 @@
...
+
+Use the properties to add to the classes of the elements containing the slots,
+to define some ids, pass Aria labels and show an optional progress bar in the
+header. You can also pass any valid properties of the `<Page>` template this
+is based on.
+  
+### Properties
+
+- `title`: The required title of the page.
+- See `FrontPage.type.ts` or code completion info.
```
jaguarfi, line 30: same comment about adding more explicit details here

frontend/src/lib/templates/page/Page.svelte
```diff
@@ -0,0 +1,195 @@
...
+
+<!-- Skip links for screen readers and keyboard users -->
+<a href="#{contentId}" class="sr-only focus:not-sr-only">{skipLinkLabel}</a>
+
+<!-- Drawer container -->
+<div {...$$restProps}>
+  <!-- NB. The Wave ARIA checker will show an error for this, but the use of both the 
+    non-hidden labels in aria-labelledby should be okay for screen readers. -->
+  <input
+    id={drawerToggleId}
```
jaguarfi, line 108: this should be aria hidden

frontend/src/lib/templates/frontPage/FrontPage.svelte
```diff
@@ -0,0 +1,113 @@
...
+
+<Page {title} {...$$restProps}>
+  <!-- Header -->
+  <slot name="header" slot="header" />
+
+  <!-- Default slot for Page starts -->
+
+  <!-- Hero image -->
+  {#if $$slots.hero}
+    <figure role="presentation" class="hero bg-[#d4dbef]">
```
jaguarfi, line 78: should this be inside some landmark? perhaps inside the `<banner>`

frontend/src/lib/templates/frontPage/FrontPage.svelte
```diff
@@ -0,0 +1,113 @@
...
+      <slot name="hero" />
+    </figure>
+  {/if}
+
+  <main class="flex flex-grow flex-col items-center justify-between pl-safelgl pr-safelgr pt-lg">
+    <!-- Title block -->
+    <hgroup class="max-w-xl py-lg text-center">
```
jaguarfi, line 85: This was an interesting reading. https://www.a11yproject.com/posts/how-to-accessible-heading-structure/#headings-in-html
image

jaguarfi (changes requested):
I was only able to review the Page and FrontPage templates, but I think there are already some thing we can start discussing

frontend/src/lib/templates/basicPage/BasicPage.svelte
```diff
@@ -0,0 +1,121 @@
...
+
+<Page {title} {...$$restProps}>
+  <!-- Header -->
+  <slot name="header" slot="header" />
+
+  <!-- Default slot for Page starts -->
+
+  <!-- Aside -->
+  {#if $$slots.aside}
+    <aside class={asideClass}>
```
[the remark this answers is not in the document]
kaljarv: True, now that I reread the `aside` specification, it's not the best one for this. The intended use for it, though, is not as a complementary title, but an additional note or tip relating to the main content. I'll change it to have a role of [note](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/note_role), which can be overriden.

I'll also wrap everything in `main`, which was the way I originally had it :D but because of this being an `aside`, I couldn't have it inside `main`…

frontend/src/lib/templates/page/parts/VoterNav.svelte
```diff
@@ -0,0 +1,40 @@
...
+`Page` template.
+
+### Usage
+
+```tsx
+<VoterNav/>
+```
+-->
+
+<Navigation slot="nav">
```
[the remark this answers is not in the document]
kaljarv: True, and actually we need to make it `display: hidden` because `aria-hidden` still leaves the links focusable.

frontend/src/lib/templates/page/Page.svelte
```diff
@@ -0,0 +1,195 @@
...
+
+<!-- Page title -->
+<svelte:head>
+  <title>{title} – {$page.data.appLabels.appTitle}</title>
+</svelte:head>
+
+<!-- Skip links for screen readers and keyboard users -->
+<a href="#{contentId}" class="sr-only focus:not-sr-only">{skipLinkLabel}</a>
+
+<!-- Drawer container -->
```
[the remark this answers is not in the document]
kaljarv: That's strange, bc on my Chrome the tab key works as supposed, but it seems that we should make this work using the arrow keys, as you suggested earlier. I'll try and fix that, but then there's the question should the menu, once open, be modal bc – or at least close automatically if the tab key is pressed… It's surprisingly involved 😅

frontend/src/lib/templates/page/Page.svelte
```diff
@@ -0,0 +1,195 @@
...
+  <input
+    id={drawerToggleId}
+    bind:checked={drawerOpen}
+    type="checkbox"
+    class="drawer-toggle"
+    tabindex="-1"
+    aria-label={drawerToggleLabel} />
+
+  <!-- Drawer content -->
+  <div class="drawer-content flex flex-col">
```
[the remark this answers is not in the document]
kaljarv: I'll do that!

frontend/src/lib/templates/page/Page.svelte
```diff
@@ -0,0 +1,195 @@
...
+        </slot>
+      </button>
+      <!-- The rest of the header -->
+      <div class="flex gap-0">
+        <slot name="header" />
+      </div>
+      {#if progress != null}
+        <meter
+          id="main-progress-bar"
+          value={progress}
```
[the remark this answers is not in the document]
kaljarv: That was the reasoning, because the progress value to show depends on what kind of task is being undertaken.

frontend/src/lib/templates/frontPage/FrontPage.svelte
kaljarv: That's very strange! It doesn't happen on my Chrome at least. It looks a lot like an issue with Svelte's refresh: are you getting some errors on the console at the same time?

frontend/src/lib/templates/page/Page.svelte
kaljarv: Good  catch!

frontend/src/lib/templates/frontPage/FrontPage.svelte
kaljarv: Now that I'm changing `Page` to wrap everything in `main`, it's gonna be within that, which should be structurally proper.

frontend/src/lib/templates/page/Page.svelte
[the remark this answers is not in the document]
kaljarv: Okay, now I updated this, which turned out be quite complicated with encountering some Svelte bugs as well.

Can you test the keyboard navigation now on your side? The main changes are:

1. When the drawer is closed, the whole navigation has `class="hidden"`, which entails `display: none`
2. When the drawer is open, the rest of the page has `inert="true"`, which renders it inaccessible by keyboard, in essence making the menu modal*
3. The button that closes the navigation is now inside the `<nav>` element produced by `Navigation`. This seems like a more logical place for it, because interaction can now nicely be limited to within the `<nav>`.

*The browser support for the pure html attribute is ~ 91%. There's a [polyfill](https://github.com/WICG/inert) available as well, which we might consider.

frontend/src/lib/templates/frontPage/FrontPage.svelte
kaljarv: I made a tentative component to handle this behaviour, please have a look at `$lib/components/headingGroup`.

I think the other option is to consider that a specific case for the page templates and split the title into two slots in `BasicPage` so that it accepts a `mainTitle` slot and an optional `preTitle` slot, which are inserted into a similar structure.

frontend/src/lib/templates/frontPage/FrontPage.svelte
jaguarfi, line 1: The candidate's front page will have the login; therefore, the Navigation menu may not be visible or may contain a restricted list of nav items. So, how can we modify the Nav Items in this template?

frontend/src/lib/templates/singleCardPage/SingleCardPage.svelte
```diff
@@ -0,0 +1,65 @@
+<script lang="ts">
+  import {Page} from '../page';
+  import type {SingleCardPageProps} from './SingleCardPage.type';
+
+  type $$Props = SingleCardPageProps;
+
+  export let cardClass: $$Props['cardClass'] = '';
+
+  // Merge restProps classes
+  $$restProps.class = `bg-base-300 ${$$restProps.class ?? ''}`;
```
jaguarfi, line 10: The base color is not needed, as the cards added on top of this template will most likely be white. I guess you want to use this template just to define some margins. am I wrong?

frontend/src/lib/templates/page/parts/CandidateNav.svelte
jaguarfi, line 1: How can we define which NavItems are rendered and which ones are not?

frontend/src/lib/templates/page/Page.svelte
jaguarfi, line 1: The skip menu is also broken here
skip-menu
the header goes outside (up) the page

jaguarfi (changes requested):
some comments!!

frontend/src/routes/_test/single/+page.svelte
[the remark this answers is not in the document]
kaljarv: Good catch! Fixed that now.

frontend/src/lib/templates/frontPage/FrontPage.svelte
kaljarv: Currently it would be up to whoever implements the `CandidateNav` template part to build that logic in there by selecting what to show by, e.g., checking whether the user is logged in. The `VoterNav` part already does this with the results.

frontend/src/lib/templates/singleCardPage/SingleCardPage.svelte
kaljarv: The base color here sets the page backround to teal so that cards with white background stand out from it. In addition to this, the template sets a proper width for the card and does some little layout tweaks.

frontend/src/lib/templates/page/parts/CandidateNav.svelte
kaljarv: That logic can be built directly into the this template part using the stores for example.

frontend/src/lib/templates/page/Page.svelte
kaljarv: Again, I can't reproduce that…

frontend/src/lib/components/headingGroup/HeadingGroup.svelte
```diff
@@ -0,0 +1,41 @@
...
+
+### Slots
+
+- default: the contents of the heading group.
+
+### Usage
+
+```tsx
+<HeadingGroup>
+  <PreTitle class="text-accent">Environment</PreTitle>
```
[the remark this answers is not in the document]
kaljarv: I can change it to `PreHeading`. The proper term for it is actually "Kicker" but I doubt many people know that :D

frontend/src/lib/components/headingGroup/HeadingGroup.svelte
```diff
@@ -0,0 +1,41 @@
...
+  <h1>The Question About Environment</h1>
```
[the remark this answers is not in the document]
kaljarv: I'll add it here, but in the other components' examples, I've held to a line not using translation strings, because they would get so cluttered with those and IMO making the examples less clear.

frontend/src/lib/i18n/en.json
```diff
@@ -1,11 +1,19 @@
 {
+  "aria": {
+    "aria.headingGroup": "Heading group",
+    "aria.preTitle": "Pre-title",
```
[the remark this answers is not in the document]
kaljarv: Oh, wasn't aware of that. I'll order this under `components.headingGroup`.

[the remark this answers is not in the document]
kaljarv: Oh, that's a typo :D Good that you pointed it out!

PR #264, opened 2023-11-18
feat: AppLogo component

kaljarv:
## WHY:

Needed as part of page templates fixing #214

The `AppLogo` component can be used to place the application's logo.
It currently handles switching between light and dark modes and offers
three predefined sizes in which the logo can be used.

In the future, the component should also allow the use of an image file
defined in Strapi.

Additionally, a `publisherName` view text is added to Strapi that is used 
in the component as the default alt text.


frontend/src/lib/components/appLogo/AppLogo.svelte
```diff
@@ -0,0 +1,62 @@
...
+</script>
+
+<!--
+@component
+Used to show the application's logo. The logo colour changes dynamically based
+on whether the light or dark mode is active.
+
+### Properties
+
+- See `AppLogo.type.ts` or code completion info.
```
jaguarfi, line 39: if we think of this as a library. it is not really helpful to say that you should see the internal files of the library 😅

frontend/src/lib/components/appLogo/AppLogo.svelte
```diff
@@ -0,0 +1,62 @@
...
+- Any valid attributes of the `<div>` element wrapping the light and dark
+  `<img>` elements
+
+### Usage
+
+```tsx
+<AppLogo size="lg"/>
+<div class="bg-primary">
+  <AppLogo inverse={true}/>
```
jaguarfi, line 48: ``
```suggestion
  <AppLogo inverse/>
```

frontend/src/lib/components/appLogo/AppLogo.svelte
```diff
@@ -0,0 +1,62 @@
...
+</div>
+```
+-->
+
+<div {...$$restProps}>
+  <img
```
jaguarfi, line 54: why are you using one  `img` on top of the other? Just curious :).

frontend/static/images/openvaa-logo-grey.svg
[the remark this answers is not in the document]
jaguarfi: I understand we want to get the file from Strapi, but we may need a backup file in the frontend, just in case.

jaguarfi (changes requested):
I understand why you want to add a wrapper for the AppLogo, but I think this is not a component as such it is more like a utility. The AppLogo is tightly linked to the application and can't really stand on its own, because it depends on the backend. Could we put this in another folder, if we put it in another folder outside of components then it would make sense to keep the logo in the `static` folder. 

We have to think that components will at some point be a library and should not depend on the application itself.

frontend/static/images/openvaa-logo-grey.svg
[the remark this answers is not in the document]
kaljarv: True :+1:, I'll move the svgs in the same folder. I just had them in static bc of the suggestion with the icons but it makes sense to be consistent with these.

frontend/src/lib/components/appLogo/AppLogo.svelte
```diff
@@ -0,0 +1,62 @@
+<script lang="ts">
+  import {page} from '$app/stores';
+  import type {AppLogoProps} from './AppLogo.type';
+
+  type $$Props = AppLogoProps;
+  export let alt: $$Props['alt'] = $page.data.appLabels.viewTexts.publisherName;
```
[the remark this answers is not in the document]
kaljarv: The reason I'm including `alt` is that the logo could be the only visible indication of the publisher of the app and it may be used outside the navigation menu as well. In contrast to icons, it's not purely decorative so IMO it should be visible to screen readers as well.

But if you feel strongly that `alt` should  be removed, I'll do that.

frontend/src/lib/components/appLogo/AppLogo.svelte
kaljarv: I'll add the descriptions from the type file like in the other components :smile:

frontend/src/lib/components/appLogo/AppLogo.svelte
kaljarv: That's because I'm using the tailwind classes to decide whether to show the light or dark version of the logo based on dark mode. Only one of them is displayed at a time.

frontend/src/lib/components/appLogo/AppLogo.svelte
```diff
@@ -0,0 +1,62 @@
...
+  export let inverse: $$Props['inverse'] = false;
```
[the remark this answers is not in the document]
jaguarfi: What is the use of the inverse prop?

frontend/src/lib/components/appLogo/AppLogo.svelte
jaguarfi: but both of them get rendered. wouldn't make more sense to add a `{#if}` instead
image

johannalaitila (approved):

frontend/src/lib/components/appLogo/AppLogo.svelte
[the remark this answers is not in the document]
kaljarv: `inverse`: If `true`, the light and dark versions of the logo will be reversed.
 Set to `true` if using the logo on a dark background.

frontend/src/lib/components/appLogo/AppLogo.svelte
kaljarv: That's surprisingly tricky, bc we need to create a media query listener for that, but it can be done. I'd suggest we use this quick version for now and create a separate issue for that, bc it'd make most sense to implement the dark mode listener as a global store, which other components can also import.

Would that be okay?

PR #266, opened 2023-11-19
feat: Show candidate photo on results & candidates page

johannalaitila:
## WHY:

Fetches candidates photo from Strapi backend and shows it in the frontend.

Fixes issue #110

Candidate photos show up in candidate browsing view if they have an uploaded photo.

image

Photo is shown also in the individual candidate page. We didn't have further designs yet in Figma so I put it in here for now.
image


frontend/src/lib/utils/constants.ts
```diff
@@ -1,4 +1,6 @@
 export const constants: Record<string, string> = {
   BACKEND_URL: import.meta.env.VITE_BACKEND_URL || process.env.VITE_BACKEND_URL || '',
+  PUBLIC_BACKEND_URL:
+    import.meta.env.VITE_PUBLIC_BACKEND_URL || process.env.VITE_PUBLIC_BACKEND_URL || '',
```
johannalaitila, line 4: TODO: Before merging this, get changes from PR #267 and implement them here properly as well to get the image urls working.

frontend/src/lib/api/getData.ts
```diff
@@ -156,6 +157,13 @@ export const getNominatedCandidates = ({
           },
           politicalExperience: attr.politicalExperience
         };
+
+        let photoURL = attr?.photo?.data?.attributes?.url;
```
kaljarv, line 161: For this `attr.photo.data?.attributes.url` should have enough safeguards.

frontend/src/lib/api/getData.ts
```diff
@@ -156,6 +157,13 @@ export const getNominatedCandidates = ({
...
+        if (photoURL) {
```
kaljarv, line 162: I guess we're going for a quick solution :smile:, but for a more permanent solution, we might want to save also the alt text, width, height and some of the different resolution (thumbnail and large, for example) data in a `photo` property.

frontend/src/lib/components/candidates/CandidatePhoto.svelte
kaljarv, line 1: Maybe we could change this to a `(entity)Avatar` component and put it in its own folder, because it'd work the same way for both parties and candidates? I'd call it 'avatar', because if there's no photo it displays the initials.

And we might want to wrap anything within it in a `<figure>` element so it'd always be semantically identical and we might also use the same kind of `restProps` delegation as in the other components.

As for sizing, I think we could provide a `size` prop with some predefined sizes (e.g. `md` and `lg`) or just use `h-X` and `w-X` classes on the containing element and use `cover` for image fitting. This way the component would always have the same size, regardless of photo dimensions or even if it's just initials.

Also, maybe just one `title` prop would be enough, and it'd be used as the `alt` text as well as generating the initials when no photo is available?

Ofc this makes the component a bit more complicated 😅 , so we can also postpone these as a separate issue for later.

frontend/src/lib/components/entityCard/EntityCard.svelte
```diff
@@ -34,15 +35,7 @@
...
-      {#if imgSrc}
-        <img class="rounded-sm" src={imgSrc} alt={imgAlt} width={imgWidth} height={imgHeight} />
-      {:else}
-        <div class="placeholder avatar">
-          <div class="w-[3.125rem] rounded-full bg-base-300">
-            <span class="text-2xl">{title.charAt(0)}</span>
-          </div>
-        </div>
-      {/if}
+      <CandidatePhoto photoURL={imgSrc} alt={imgAlt} width={imgWidth} height={imgHeight} {title} />
```
kaljarv, line 38: We can remove `<svelte:fragment>` and add ` slot="card-media"` directly to `CandidatePhoto`.

If we pass the dims, they should be just `{imgWidth} {imgHeight}` but as we don't actually have those, maybe just leave them out now?

frontend/src/lib/api/getData.ts
kaljarv: At the same time, we should change the Strapi data def starting from line 172 in `getData.type.ts` thus:

```ts
// inside StrapiCandidateData.attributes
    photo: {
      data?: {
        id: number | string;
        attributes: StrapiImageData;
      };
    };
    politicalExperience: string;
  };
}

export interface StrapiImageData extends StrapiImageFormatData {
  alternativeText?: string;
  caption?: string;
  formats: {
    thumbnail?: StrapiImageFormatData;
    small?: StrapiImageFormatData;
    medium?: StrapiImageFormatData;
    large?: StrapiImageFormatData;
  };
}

export interface StrapiImageFormatData {
  ext: string;
  height?: number;
  name: string;
  size?: number; 
  url: string;
  width?: number;
}

```

frontend/src/lib/api/getData.ts
johannalaitila: Yeah sure that could be done, we didn't have any plans when I started creating this but I'll add these to the PR!

kaljarv (approved):
LGTM!

Just one thing: Could we maybe leave the property name as just `photo` instead of `photoURL` for brevity?

FYI: I'm refactoring the candidate components in PR #425 but if we merge this before that I can rebase. There's also an `Avatar` component that will replace `CandidatePhoto`.

PR #283, opened 2023-11-30
Add generic question types for non-opinion details

### LINKED ISSUE #194: Enable generic question types for non-opinion details.
Use the `Question`-`Answer` model for other details as well, such as gender (an enumerated type) and election manifesto (just text).

We may also need a way to distinguish opinion questions from these non-opinion ones. This can be done either via `Question Types` or as a boolean in `Question Categories`.

joonatan-aatos:
## WHY:

Fixes #194 


frontend/src/routes/candidate/profile/+page.svelte
anolpe, line 1: This file should be removed because it has been moved to candidate/(protected)/profile

anolpe (changes requested):
Voter app is broken and doesn't give any results after answering questions.

Issue #278, opened 2023-11-26
Clicking on skip link breaks layout


kaljarv:
Investigate and exterminate the bug with the skip link on the `Page` template, detailed in [comment A](https://github.com/OpenVAA/voting-advice-application/pull/258#discussion_r1403810962) and [comment B](https://github.com/OpenVAA/voting-advice-application/pull/258#discussion_r1405272390) in pull #258.

kaljarv:
@jaguarfi, can you share which browser setup did you have when encountering the errors?

PR #303, opened 2023-12-14
fix: reactivity issues in components and templates

kaljarv:
## WHY:

Fixes: #278 

Components and templates using `$$restProps` forwarding had an issue with reactive
updates that would lead to layouts breaking. The reason for this was that the
`class` string used by the components were merged into `$$restProps` in a
non-reactive part of the component and the `$$restProps` was passed
directly to the component with the expectation that it would contain the proper
`class` string. However, any reactive updates would reinstantialize 
`$$restProps` and the alterations would no longer be included in it.

This issue became very prominent with the introduction of page templates,
because any reactive update to a page's contents, would break the layout
due to the loss of  `class="drawer"` in the `<div>` containing the page,
which was also described in #278.

In this update, default `class` strings are recombined with `$$restProps` on
reactive updates using a shared utility function (`concatClass`/`concatProps`).
In addition, the effects of some other styling-related custom properties of 
components, such as `size` and `color`, are made reactive as well.

Also includes a quick styling fix to the nav menu on `Page`, which is not
related to reactivity issues.


anolpe (approved):
LGTM!

PR #324, opened 2024-01-05
build: fix Eslint and Prettier conflicts

### LINKED ISSUE #323: Eslint and Prettier conflicts make some checks always fail
See run #617 on PR #317.

Here, the Eslint's and Prettier's opinions on indenting the ternary operator always cause either check to fail.

Running on the automatic check `npx eslint-config-prettier .eslintrc.json` reports:

```
The following rules are unnecessary or might conflict with Prettier:

- indent
- linebreak-style
- semi

The following rules are enabled but cannot be automatically checked. See:
https://github.com/prettier/eslint-config-prettier#special-rules

- quotes
```

This is due to the fact that `eslint-config-prettier` cannot automatically turn off rules, only defaults.

To fix, the first three rules should be removed, and the `quotes` one edited and kept in sync with Prettier:

Eslint:

```
{
  "rules": {
    "quotes": [
      "error",
      "single",
      { "avoidEscape": true, "allowTemplateLiterals": false }
    ]
  }
}
```

Prettier:

```
{
  "singleQuote": true
}
```

kaljarv:
## WHY:

Fixes #323 

Conflicts Eslint's and Prettier's opinions cause some checks always to fail.

Running on the automatic check `npx eslint-config-prettier ./frontend/.eslintrc.json` reports:

```
The following rules are unnecessary or might conflict with Prettier:

- indent
- linebreak-style
- semi

The following rules are enabled but cannot be automatically checked. See:
https://github.com/prettier/eslint-config-prettier#special-rules

- quotes
```

This is due to the fact that `eslint-config-prettier` cannot automatically turn off rules, only defaults.

To fix, the first three rules need to be removed, and the `quotes` one edited and kept in sync with Prettier:

Eslint:

```
{
  "rules": {
    "quotes": [
      "error",
      "single",
      { "avoidEscape": true, "allowTemplateLiterals": false }
    ]
  }
}
```

Prettier:

```
{
  "singleQuote": true
}
```

In the backend config, the Eslint config also needs to `extend` `prettier`.

### DIFF
--- a/backend/vaa-strapi/.eslintrc.json
+++ b/backend/vaa-strapi/.eslintrc.json
@@ -1,6 +1,6 @@
 {
   "parser": "@typescript-eslint/parser",
-  "extends": "eslint:recommended",
+  "extends": ["eslint:recommended", "prettier"],
   "env": {
     "commonjs": true,
     "es2017": true,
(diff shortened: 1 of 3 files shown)


johannalaitila (approved):
LGTM

Issue #280, opened 2023-11-26
Convert the Voter App routes to use page templates


kaljarv:
Now that the page templates are ready, the Voter App should be converted to use them.

At the same time, the `appType` store should be updated on the Voter App side as well.

Issue #198, opened 2023-09-25
As a Candidate, I want to enter my opinions to the multiple choice statements.


kaljarv:
The candidate should be able to enter their multiple choice opinions to Likert questions.

The display of the choices should be visually as close to the voter's view of the same questions as possible:
- The same Svelte components should be used if possible.
- The candidate should also be able to see the same information about the question that is offered to the voter.

The candidate should see the metadata for the question:
- the `Question Category` it belongs to
- any `Election`, `Constituency` or `Entity Type` the question is restricted to

Note that other types of questions will be implemented in the future, so the model should be flexible.

Issue #109, opened 2023-06-03
As a Voter, I want to see the candidates' answers to the questions.


jaguarfi:
This component is not in Figma yet.

PR #333, opened 2024-01-14
feat: question subcomponents

kaljarv:
## WHY:

Creates a suite of components for consistent display of questions for
both voters and candidates as well as for answering and presentation
of already given answers.

- Needed for #109 and #198 
- Partially solves #280

Creates several components for displaying different parts of
a question, such as `QuestionActions`, which displays the additional
actions available, such as 'Next', 'Previous' and 'Delete answer'.

The purpose for providing these partial components is to show the
questions consistently for both voters and candidates when answering
because any differences in how they're displayed may effect answers.

Furthermore, the `LikertResponseButtons` now supports use both for
answering (for both the voter and a candidate) and for displaying
answers when a voter is looking at a candidate's profile.

The `/(voters)/questions` route is update to show the components in
action. They are also quickly added to the `CandidateOpinions`
component.

Furthermore, there is a preliminary update to the 
`/candidate/(protected)/questions` route because the `Question`
component is no longer available.

Finally, there are small updates to global types, `Expander` styling
and a `onKeyboardFocusOut` Svelte action.


jaguarfi (commented):
LGTM

frontend/src/lib/components/expander/Expander.svelte
```diff
@@ -66,47 +60,43 @@
...
+
+  <Expander title="Example title" variant="category"  iconColor="primary" 
+    titleClass="bg-base-100 text-primary" contentClass="bg-base-300 text-info font-bold">
+    <p>Example content<p/>
+  </Expander>
+  ```
+-->
 
 <div class={collapseClasses}>
   <input type="checkbox" aria-label="open ${title}" on:click={toggleExpanded} />
```
jaguarfi, line 102: What about the `aria-label` when the expander is supposed to close? Maybe we need to update the label dynamically.

frontend/src/lib/components/questions/LikertResponseButtons.svelte
```diff
@@ -0,0 +1,270 @@
...
+  options={question.options}
+  mode="display"
+  selectedKey={$voterAnswers[question.id]}
+  entityKey={candidateAnswer}
+  entityLabel={$_('candidate.candidateAnswerLabel')}
+   />
+```
+-->
+
+<fieldset
```
jaguarfi, line 203: If users want to use the keyboard, how should they navigate the answers? Will they use arrow keys or the Tab key?

frontend/src/lib/utils/onKeyboardFocusOut.ts
```diff
@@ -0,0 +1,44 @@
...
+ *   has left the container. We listen to `focusout` and `focusin` to
+ *   track whether we have focus or not. (We cannot use
+ *   `document.activeElement` because it's not reliable.)
+ *
+ * @param node The target element
+ * @param callback The event handler to call
+ *
+ * @example `<Navigation on:keyboardFocusOut={closeDrawer} />`
+ */
+export function onKeyboardFocusOut(node: HTMLElement, callback: () => void) {
```
jaguarfi, line 23: This looks a bit hacky. We can revisit this in the future and find other alternatives.

frontend/src/lib/components/expander/Expander.svelte
kaljarv: We should actually get rid of the label altogether and use this [pattern](https://www.w3.org/WAI/ARIA/apg/patterns/disclosure/examples/disclosure-faq/). We had a discussion about that in the Candidate App PR that introduced `Expander`, but it was left unresolved.

I made an issue out of it #341

frontend/src/lib/components/questions/LikertResponseButtons.svelte
kaljarv: They will use the arrow keys, which is the default way a radio group operates. `Tab` will move focus out of the widget, which is also the default.

frontend/src/lib/utils/onKeyboardFocusOut.ts
kaljarv: I agree, it's quite contrived. I spent quite some time trying come up with a nicer solution, but this was the only working one I could find. One tricky issue is that when tabbing through the navigation menu, keyboard focus will move out of the page after you reach the last nav item. Thus, we won't have any focused element on the page anymore…

PR #343, opened 2024-01-29
refactor: Shared module and password-related changes

TrixterTheTux:
## WHY:

Refactor the password validation to use a shared module instead to avoid having to define password validation rules in two different places. Additionally, the password reset and change endpoints now also enforce on the backend that the password matches the requirements.

There was also some discussion about wanting to have the email templates outside the code itself, so `backend/vaa-strapi/config/email-templates/` has been added with respective documentation so that it could be easily edited.

Some of the Docker-related files have been refactored so that the context of the build happens in the root of the project instead of their respective subfolders. This is required so that the shared module can be added, as Docker does not allow you to copy files outside the build root. 

For sharing the module between the components itself it uses a symlink to the shared module so that it works locally and inside the Docker container without additional trickery. I tried some other approaches but they eventually weren't ideal:
- TS supports Project References, but this requires changing the `rootDir` to include the shared module as the root. This would otherwise be ideal, but Strapi depends on the rootDir being the Strapi dir and I could not find a way to be able to set the path where Strapi would reside, as rootDir affects the build output (e.g. Strapi ends up being in ./dist/backend/vaa-strapi/ but tries to load ./dist/ directly).
- yarn links. As the project currently supports both `npm` and `yarn`, it would require dropping `npm` completely/additional dependencies meaning this wouldn't be supported out of the box.


backend/vaa-strapi/src/extensions/users-permissions/strapi-server.js
```diff
@@ -59,16 +66,11 @@ module.exports = async (plugin) => {
...
-    // https://github.com/strapi/strapi/blob/2a2faea1d49c0d84077f66a57b3b73021a4c3ba7/packages/plugins/users-permissions/server/bootstrap/index.js#L41-L79
-    email.reset_password.options.message = `<p>We heard that you lost your password. Sorry about that!</p>
-
-<p>But don’t worry! You can use the following link to reset your password:</p>
-<a href="<%= URL %>?code=<%= TOKEN %>"><%= URL %>?code=<%= TOKEN %></a>
-
-<p>Thanks.</p>`;
+    email.reset_password.options.message = fs
+      .readFileSync('config/email-templates/reset-password.html')
+      .toString();
```
jaguarfi, line 73: So, is this change intended to solve the clickable URL problem?

jaguarfi (commented):
Could it be possible to create another docker composable to manage the shared utilities? As the utilities will be used for both the backend and frontend, it would make sense to have an independent composable with a different building process. (Maybe at some point it could be replaced with our own npm package)

backend/vaa-strapi/src/extensions/users-permissions/strapi-server.js
TrixterTheTux: One of the goals, yes. Additionally, this allows easy modification of the template to the application's administrator as requested during a sprint review (compared to being hardcoded in the code).

TrixterTheTux:
> Could it be possible to create another docker composable to manage the shared utilities? As the utilities will be used for both the backend and frontend, it would make sense to have an independent composable with a different building process. (Maybe at some point it could be replaced with our own npm package)

Could you elaborate on how exactly this would work? The idea with the current design is that the shared module isn't directly an additional dependency to avoid having to package it to npm/manage to build it separately from the frontend/backend for every change made to it.

jaguarfi:
> > Could it be possible to create another docker composable to manage the shared utilities? As the utilities will be used for both the backend and frontend, it would make sense to have an independent composable with a different building process. (Maybe at some point it could be replaced with our own npm package)
> 
> Could you elaborate on how exactly this would work? The idea with the current design is that the shared module isn't directly an additional dependency to avoid having to package it to npm/manage to build it separately from the frontend/backend for every change made to it.

I was thinking that we could create a docker container for the shared code and add it to the docker-compose.

joonatanaatos (approved):

joonatanaatos (approved):

joonatanaatos (approved):

samuliys (approved):

PR #390, opened 2024-02-27
build(deps)!: update to DaisyUI 4.7.2

### LINKED ISSUE #256: Update DaisyUI to version 4
There's a cpla [breaking changes](https://github.com/saadeghi/daisyui/blob/master/CHANGELOG.md#changelog) that need to fixed at the same time.

1. Change all direct CSS vars to use `oklch(var(--p))` instead of `hsl()`
2. `--btn-text-case` CSS var is no longer used and can be removed from `tailwind.config.cjs`

At the same time, also check that `modal` works properly. See [discussion](https://github.com/OpenVAA/voting-advice-application/pull/316#discussion_r1443812843).

kaljarv:
## WHY:

Fixes #256 

Also, make necessary fixes to `tailwind.config.cjs` and the `ScoreGauge` and `LikertResponseButtons` components that use the DaisyUI color CSS vars (see below).

BREAKING CHANGE: `hsl(var(--n))` and similar CSS var calls have to be changed to `oklch(var(--n))`.

See also:
https://github.com/saadeghi/daisyui/blob/master/CHANGELOG.md#400-2023-11-12i

Includes also a quick fix to the `OpenVAALogo` component.

### DIFF
--- a/frontend/src/lib/components/openVAALogo/OpenVAALogo.svelte
+++ b/frontend/src/lib/components/openVAALogo/OpenVAALogo.svelte
@@ -1,4 +1,5 @@
 <script lang="ts">
+  import {concatClass} from '$lib/utils/components';
   import type {OpenVAALogoProps} from './OpenVAALogo.type';
 
   type $$Props = OpenVAALogoProps;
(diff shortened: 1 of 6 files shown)


frontend/src/lib/components/openVAALogo/OpenVAALogo.svelte
```diff
@@ -9,28 +10,27 @@
...
-    case 'sm':
-      classes = 'h-20 pt-2';
-      break;
-    case 'lg':
-      classes = 'h-32 pt-6';
-      break;
-    default:
-      classes = 'h-24 pt-4';
-  }
+  $: {
```
jaguarfi, line 13: What is the functionality of the `$:`?

kaljarv: If `classes` is not built within a reactive block, any changes to the `size` attribute after component creation won't have any effect. One use case where this could be needed is if the logo is used in a sticky menu bar that shrinks when scrolled down.

PR #395, opened 2024-02-29
feat: enable party and category colors, question ordering and create local a file for local app instance settings

### LINKED ISSUE #35: Store settings values in backend

### LINKED ISSUE #173: Fix question ordering
Fetcth types, then categories ordered by order with questions populated. Or the other way round.

At the same time, also separate the info questions into groups in `<CandidateBasicInfo>`.

### LINKED ISSUE #374: feat: enable colours for Parties
Categories need to have a theme colour,  this can be added as a `color` property of the `Question Category` objects. `Parties` already have this, but it has no effect in the frontend.

The color should be applied to the category term when shown next to the questions and possibly in category headings (in the Candidate App). This implies changes to the frontend data types, such that the `category` of a `QuestionProp` is an object instead of just a string containing the name.

For parties, the colour should be applied to the party flag icon. This, in turn, would possibly imply changes to the `Icon` component to allow for a custom color in addition to the named ones. This would differ from the so that the it would:
- either automatically create a different version for use in dark mode
- accept to colors, `customColorLight` and `customColorDark` for use in light and dark modes

Finally, before using any of the colors in the frontend, the colour should be automatically modified to have sufficient contrast on `base-1/300`. Possibly:
- use the formula for contrast calculation in this [discussion](https://stackoverflow.com/questions/65304462/is-it-possible-to-make-a-javascript-program-that-changes-the-color-of-the-text-t)
- either lighten or darken by mixing with black or white and either solve for correct color analytically or iteratively ([example here](https://medium.com/confrere/automatically-creating-an-accessible-color-palette-from-any-color-sure-e735c3f2f45e), although it uses a dep)

### LINKED ISSUE #396: feat: colors for Question Categories
See #374

Implement at the same time as the required content model changes: #389

At the same time, update the `SubMatches` component as well.

kaljarv:
## WHY:

Fixes #374
Fixes #35
Fixes #396 
Fixes #173 

Implements party and question category colors in the frontend. The colors are also automatically adjusted to reach AA-level color contrast on both light and dark backgrounds.

Introduces a `PartyTag` component that can be used to show the party name and an icon in the correct color as well as a `CategoryTag` component for showing question category names in the right color.

For the contrast checking to work, we also need access to the color definitions, which are now moved from `tailwind.config.cjs` to `$lib/config/settings.json`, which can be used to easily set instance-specific options. Supported locale settings are also migrated to that file.

The settings are also accessible from a `settings` store that can later on be update to include effective settings merging any choices made by the app user to the default ones.

Also implements question ordering based on the `order` properties of questions and categories.


kaljarv:
Apologies for any reviewers that this is a bit bloated but all of these changes are related…

PR #421, opened 2024-03-11
fix: locale switching with server-loaded data

### LINKED ISSUE #420: fix: localised text is not updated when it is server-loaded as part of data and the locale is changed multiple times
When data is loaded on the server and updated due to locale changes, components using it are not automatically rerendered by the client.

kaljarv:
## WHY:

Fixes #420 

Fixes localised text updates when the localised data is part of server-loaded data, including `App Labels`.

Server `load` functions now depend on either the `lang` route param or the outermost layout's `data.i18n.currentLocale`, which is awaited for using `parent()`.

When data is loaded on the server and updated due to locale changes, components using it are not automatically rerendered by the client. This update makes all such data reactive in the frontend.

NB. The subobjects of `data` used are now made explicit, but a more concise way would be to remove `export let data` altogether and always call `$page.data.foo` to get the requisite data.

### DIFF
--- a/backend/vaa-strapi/src/components/labels/action-labels.json
+++ b/backend/vaa-strapi/src/components/labels/action-labels.json
@@ -49,7 +49,7 @@
     "answerCategoryQuestions": {
       "type": "string",
       "required": true,
-      "default": "Answer {{0}} Questions"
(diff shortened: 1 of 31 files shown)


johannalaitila (approved):
LGTM

PR #433, opened 2024-03-20
feat: admins can send registration emails to candidates

### LINKED ISSUE #183: As an Application Admin, I want to send a login email to all candidates.
The email should contain an introduction and a personal link that allows the candidate to login to the app and set their password.

The link is connected to a specific `Candidate` or `User`.

The introductory message should be defined by the Application Admin when sending the message.

Because the message may be multilingual, the link needs to be at an understandable location. If the link uses a label different from the url, the label needs to be specifiable by the Application Admin.

The UI for the Application Admin tools does not need to be mobile-friendly.

anolpe:
## WHY:

Fixes #183 

Functionality to send registration emails to unregistered candidates was added to Strapi admin panel.

image

It is also possible to send registration email to single candidate.

image

- [x] Is there an existing issue linked to this PR? #183 


joonatanaatos (approved):

backend/vaa-strapi/src/api/admin/services/admin.ts
```diff
@@ -0,0 +1,54 @@
...
+      text
+    });
+  },
+  sendEmailToUnregistered: async (subject: string, content: string) => {
+    const unregisteredCandidates = (await strapi.entityService.findMany(
+      'api::candidate.candidate',
+      {
+        fields: ['registrationKey', 'email'],
+        filters: {
+          registrationKey: {
```
TrixterTheTux, line 38: I feel like checking for `user` relation would be more reliable indicator if they're unregistered.

TrixterTheTux (changes requested):
I will also do some more investigation regarding the proper handling of permissions on the admin side.

TrixterTheTux (approved):

PR #437, opened 2024-03-22
Feat 409: Shows a warning if a user is logged in when registering

Henriki2305:
## WHY:

Fixes [#409 ]

Adds a warning that shows if a user is currently logged in

image

- [x] Is there an existing issue linked to this PR? #409 


samuliys (changes requested):
The user usually registers using the link, which skips this page completely. Therefore, I think that this warning should also be added to the page where the candidate sets their password for the first time.

In addition, the warning states that the user should log out, but there is no logout button visible on the page. Should it be visible?

frontend/src/lib/templates/frontPage/FrontPage.svelte
```diff
@@ -48,6 +48,7 @@ is based on.
     <h1>{$page.data.election.name}</h1>
   </svelte:fragment>
 
+
```
joonatanaatos, line 51: Unnecessary newlines shoud be eschewed

frontend/src/lib/candidate/components/PasswordField/PasswordField.svelte
```diff
@@ -4,6 +4,7 @@
   export let password = '';
   export let autocomplete = '';
   export let id = 'password';
+  $: fieldId = id;
```
joonatanaatos, line 7: What is the purpose of this? I'd assume Svelte automatically re-renders when the props change. If there is a good reason, an inline comment should be added

samuliys (approved):
LGTM after addressing the `fieldId` situation

PR #451, opened 2024-03-30
fix: relocate logout button below the warning on registration page

Henriki2305:
## WHY:

relocates the button for logging out when a user is logged in when trying to register below the warning

image
image

- [X] I have reviewed the changes myself in this PR. Please check the [self-review document](https://github.com/OpenVAA/voting-advice-application/blob/main/docs/contributing/self-review.md)
- [X] I have run the unit tests successfully.
- [X] I have run the e2e tests successfully.
- [X] I have tested this change on my own device.
- [X] I have tested this change on other devices (Using Browserstack is recommended).
- [X] I have tested my changes using the [WAVE extension](https://wave.webaim.org/extension/)
- [X] I have added documentation where necessary.
- [X] Is there an existing issue linked to this PR?


anolpe (approved):
LGTM!

frontend/src/lib/candidate/components/logoutButton/LogoutButton.svelte
```diff
@@ -78,7 +82,11 @@ When set to false, the button variant is ghost.
     text={$t('candidateApp.navbar.logOut')}
     color="warning" />
 {:else}
-  <Button on:click={triggerLogout} text={$t('candidateApp.homePage.logOut')} color="warning" />
+  <Button
+    on:click={triggerLogout}
+    text={$t('candidateApp.homePage.logOut')}
+    color="warning"
+    variant="main" />
```
samuliys, line 89: This component is also used, for example, in the front page, which now looks like this:
image

In the figma designs the style of the front page log out button is different so setting the variant to "main" here should probably be done in a different way. One fix would be that the variant could be given as props to the LogoutButton.

samuliys (changes requested):

frontend/src/lib/candidate/components/logoutButton/LogoutButton.svelte
```diff
@@ -7,8 +7,10 @@
...
 
-  /** Defaults to true, so that button variant is "icon". Can be set to false if the button should be of variant "main" */
-  export let variantIcon = true;
+  /** Defaults to normal, button variant can be "normal" for just text, "icon" for a button with an icon and "main" for a filled red box */
+  export let buttonVariant = 'normal';
+  /*Determines whether the user is taken to the login page after logging out or stays on the current page*/
+  export let stayOnPage = false;
 
   /** time until automatic logout for modal */
   export let logoutModalTimer = 30;
```
samuliys, line 16: As the component has become more complex I'd suggest that a `LogoutButton.type.ts` file is created with the type definitions and explanations, i.e.
```
export type LogoutButtonProps = {
  /**
   * Some text
   *
   * @default 'normal'
   */
  buttonVariant?: 'normal' | 'icon' | 'main';
  /**
   * ...
   *
   * @default false
   */
  stayOnPage?: boolean;
  /**
   * ...
   *
   * @default 30
   */
  logoutModalTimer?: number;
};
```
and then those are used instead in the component
```
import type {LogoutButtonProps} from './LogoutButton.type';

type $$Props = LogoutButtonProps;
export let buttonVariant: $$Props['buttonVariant'] = 'normal';
...
```
and so on

frontend/src/lib/candidate/components/logoutButton/LogoutButton.svelte
```diff
@@ -59,24 +63,31 @@
...
+This component has optional string property `buttonVariant`:
+When set to normal (default), the button variant is normal which is just plain text. 
+When set to icon, the button variant is icon.
+When set to main, the button variant is main which is a red box.
 
 ### Usage
 ```tsx
   <LogoutButton />
 ```
 -->
```
samuliys, line 75: There are also missing props in the component documentation so those should be added.
Also, the component documentation could be improved to be more in the style of the other components where the properties are clearly documented. Here is an example regarding how the properties should be documented to get a better idea:
```
### Properties

- `name`: the name of the icon to use
- `size`: The size of the icon as one of the predefined sizes 'sm', 'md' or 'lg'.
   For arbitrary values, you can supply a `class` property, such as 
   `h-[3.15rem] w-[3.15rem]`. @default 'md'
- `color`: The color of the icon as one of the predefined colours.
   For arbitrary values, you can supply a `class` property, such as
   `fill-[#123456]`. @default 'current'
- `aria-hidden`: @default `true`
- `role`: Aria role @default `img`
- `class`: Additional class string to append to the element's default classes.
```

(from `Icon.svelte`)

frontend/src/routes/[[lang=locale]]/candidate/(protected)/+page.svelte
```diff
@@ -138,6 +138,6 @@
         href={nextAction.href} />
     {/if}
 
-    <LogoutButton variantIcon={false}></LogoutButton>
+    <LogoutButton></LogoutButton>
```
Comment (samuliys) [inline frontend/src/routes/[[lang=locale]]/candidate/(protected)/+page.svelte:141]: Should be simplified to just `<LogoutButton />`

samuliys (changes requested):

frontend/src/lib/candidate/components/logoutButton/LogoutButton.svelte
```diff
@@ -6,12 +6,16 @@
...
   import {getContext} from 'svelte';
   import type {CandidateContext} from '$lib/utils/candidateStore';
+  import type {LogoutButtonProps} from '$lib/types/LogoutButton.type';
 
-  /** Defaults to true, so that button variant is "icon". Can be set to false if the button should be of variant "main" */
-  export let variantIcon = true;
+  type $$props = LogoutButtonProps;
+  export let buttonVariant: $$props['buttonVariant'] = 'normal';
+  /** Defaults to normal, button variant can be "normal" for just text, "icon" for a button with an icon and "main" for a filled red box */
+  /*Determines whether the user is taken to the login page after logging out or stays on the current page*/
```
samuliys, line 14: These can now be removed as they are already in the type file

frontend/src/lib/candidate/components/logoutButton/LogoutButton.svelte
```diff
@@ -6,12 +6,16 @@
...
+  export let stayOnPage: $$props['stayOnPage'] = false;
 
   /** time until automatic logout for modal */
```
samuliys, line 17: Same with this

samuliys (approved):
LGTM after the minor changes

anolpe (approved):

PR #572, opened 2024-06-20
refactor: results tab persists state with snapshot

### LINKED ISSUE #568: refactor: convert results tab persisting to use Snapshot
I tested this by adding following code to the start of `(voters)/results/+page.svelte`, removing import for `activeTab` store and changing every `$activeTab` to `activeTab`.
```ts
import type {Snapshot} from './$types';

let activeTab = 0;

export const snapshot: Snapshot<number> = {
  capture: () => activeTab,
  restore: (value) => (activeTab = value)
};
```

The biggest difference compared to the store is that the active tab is persisted also when the page is refreshed. Otherwise it seems to be functionally same as the store. Visually there are two differences:

- When coming back from viewing party Candidates tab is very briefly shown as active tab
- After refreshing the first tab is shown as active tab until the data has been loaded

Of course it may be possible to fix these as this was only a quick test.

anolpe:
## WHY:

Fixes #568 

Results tab is persisted with snapshot instead of store.

### DIFF
--- a/frontend/src/routes/[[lang=locale]]/(voters)/results/+page.svelte
+++ b/frontend/src/routes/[[lang=locale]]/(voters)/results/+page.svelte
@@ -1,5 +1,6 @@
 <script lang="ts">
   import {error} from '@sveltejs/kit';
+  import type {Snapshot} from './$types';
   import {t} from '$lib/i18n';
   import {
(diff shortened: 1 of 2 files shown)


kaljarv (approved):
LGTM

PR #577, opened 2024-07-01
refactor: settings

### LINKED ISSUE #496: refactor: settings
Settings handling is a bit ad hoc and should be refactored:

### Optional and default settings

All of the settings in `settings.json` are currently required. These should be converted to optional and default values put in place when loading them. Also, consider changing the format from `json` to `ts`, so we could use shared typing for the config and effective settings, which are now duplicated in the json schema and `global.d.ts`.

### Local and fetched settings

Effective settings are now derived from the `page` store so that they can incorporate `App Settings` fetched from the backend. This leads to problems if used in SSR and Sveltekit will promptly throw [an error](https://kit.svelte.dev/docs/state-management#avoid-shared-state-on-the-server). For this reason, the local settings from the json file are directly used, whereas elsewhere the `settings` store is used.

It'd be best to always use the combined settings and if this is not possible explicitly divide the settings into static (or local) and dynamic settings so that these are not confused.

### See also

- #432
- #459

anolpe:
## WHY:

Fixes #496 

### What has been changed

- All settings can now be changed in Strapi in addition to editing the local `ts` files. A new `App Settings` content type is used for this.
- Settings are moved into the `shared` module.
- All translations can be overridden in Strapi. Translations under the `dynamic` key are inserted there by default. A new `App Customization` single type is used for this. It includes all other instance-specific customization, including logos and translation overrides
- The deprecated `App Labels` content type is removed.


shared/settings/settings.type.ts
```diff
@@ -0,0 +1,318 @@
+export type AllSettings = StaticSettings & DynamicSettings;
+
+export type StaticSettings = {
```
kaljarv, line 3: We could add general descriptions for these types:

- `StaticSettings`: These settings can only be set by editing the `staticSettings.ts` file and they cannot, thus, change after initialisation.
- `DynamicSettings`: These settings can be set either by editing the `dynamicSettings.ts` file or overwritten with settings loaded by the `DataProvider`.

Also, the `AppSettings` type should be updated to be a combination of these two.

Could you also check:

- Is there a way to make the static settings readonly so they cannot be overwritten by data loaded by accident? If not, just typing the incoming data as `DynamicSettings` ought to be pretty robust.
- Is there a use case and an easy way to copy the current settings files as defaults and then add another pair of files where the contents are `Partial`s of the settings and are gitignored (with example files provided)? The cascade would thus be:

1. Defaults (ts-files): `Static/DynamicSettings`
2. Local overrides (ts-files based on example files, optional, gitignored): `Partial<Static/DynamicSettings>`
3. DataProvider-provided overrides (fetched, optional): `Partial<DynamicSettings>`

kaljarv: Note that there might be a discrepancy with this system when adding dynamic settings if their defaults are based on the defaul local files. Maybe use the merged defaults + local overrides to set their default values?

kaljarv:
The editing of the settings works nicely in Strapi. Two things would be nice to add:

- Make all of the settings optional such that only those that are set will override the locally defined or default ones
- Add (preferably somewhat automatically) descriptions for the fields matching the comments in type definitions like this: 
image

anolpe:
> Add (preferably somewhat automatically) descriptions for the fields matching the comments in type definitions like this:
> 
> image

This isn't possible to do because descriptions are [part of layout configuration which is not saved to the model schema](https://docs.strapi.io/dev-docs/faq#can-i-store-my-content-manager-layout-configurations-in-the-model-settings).

kaljarv:
> > Add (preferably somewhat automatically) descriptions for the fields matching the comments in type definitions like this:
> > image
> 
> This isn't possible to do because descriptions are [part of layout configuration which is not saved to the model schema](https://docs.strapi.io/dev-docs/faq#can-i-store-my-content-manager-layout-configurations-in-the-model-settings).

Ok, in that case, we'll just need to refer the users to the documentation.

shared/settings/settings.type.ts
kaljarv: Let's add an empty local override file for now.

kaljarv:
Some stuff to do or consider:
- A new content type for customising the app, which will house:
  - Dynamic translations (list provided soon)
  - Publisher name, logo, poster images etc. which currently reside in `AppSettings`
  - Any arbitrary translation overrides (repeatable component)
  - In the future, illustrations for content pages to replace the emojis, such as on the intro page for example
  - FAQ items for the Candidate App as a repeateable component list
- For images, a component may be needed that allows for definition of an optional dark mode image variant

kaljarv:
The dynamic translations are now separated in #593. See `/frontend/src/lib/i18n/translations/en/dynamic.json` for details.

shared/settings/settings.type.ts
anolpe: I have worked on making settings in local override files and Strapi deeply `Partial`. One problem I encountered is that if `survey` hasn't been set in `defaultDynamicSettings.ts`, user could have following local override file:
```ts
// dynamicSettings.ts
export const dynamicSettings: DeepPartial<DynamicSettings> = {
    survey: {
        linkTemplate: "example"
    }
};
```
If this would be merged with default setting (with some generic deep merge function), resulting settings would have `survey` without required property `showIn`. So we would need to have some kind of more specialized and complex merge function.

It feels like these settings are getting pretty complicated. Is there really use case for these local orverride files? Wouldn't it work if we would have one file for dynamic settings and one file for static settings. Those files would be similar to default settings files in this branch, but user could modify those if needed. In Strapi, dynamic settings would be required and optional in the same way as in the settings type. Contents of settings file would be loaded into Strapi so user would see them and could modify even a single setting in there. Would there be any problems with this kind of approach?

kaljarv:
One layout bug that could be fixed at the same time, is the hide/show password icon overlapping the contents:

image

The lock icon, on the other hand, works as it should:

image

anolpe:
Would something like this work for `results.cardContents.candidate|party`? User would choose "submatches" or "question" (or "candidates" for party) for value of `content` and in case of "question" user would also fill other fields. I think this would be easier for user than JSON field but it's still little confusing as we can't set those additional fields to be conditionally required when user has chosen "question".

image

kaljarv:
> Would something like this work for `results.cardContents.candidate|party`? User would choose "submatches" or "question" (or "candidates" for party) for value of `content` and in case of "question" user would also fill other fields. I think this would be easier for user than JSON field but it's still little confusing as we can't set those additional fields to be conditionally required when user has chosen "question".
> 
> image

Yeah, that looks like a good compromise!

backend/vaa-strapi/src/api/app-customization/controllers/app-customization.ts
```diff
@@ -0,0 +1,74 @@
+/**
+ * app-customization controller
+ */
+
+import {factories} from '@strapi/strapi';
+import {API} from '../../../functions/utils/api';
+import {staticSettings} from 'vaa-shared';
+
+export default factories.createCoreController(API.AppCustomization, () => ({
```
kaljarv, line 9: Translations and candidate app FAQ items are now preprocessed by the backend. This could also be done in the frontend but as the response can be cached, we can save resources by doing it here.

backend/vaa-strapi/src/api/app-setting/content-types/app-setting/schema.json
```diff
@@ -12,51 +12,54 @@
...
-    },
     "underMaintenance": {
       "type": "boolean",
       "default": false
     },
     "allowOverwrite": {
       "type": "boolean",
       "default": false
+    },
+    "header": {
```
kaljarv, line 23: All the settings sections are (repeatable) components so that we get a similar hierarchical structure for them as in the frontend. Their contents are also sometimes complex, which can only be acchieved with components.

backend/vaa-strapi/src/functions/generateMockData.ts
```diff
@@ -270,54 +296,7 @@ async function createLanguages() {
   });
 }
 
-async function createElectionAppLabel() {
```
kaljarv, line 273: `App Labels` are deprecated and all references to them are removed.

backend/vaa-strapi/src/functions/utils/appCustomization.ts
```diff
@@ -0,0 +1,61 @@
+import {staticSettings} from 'vaa-shared';
+import en from './translations/en/dynamic.json';
```
kaljarv, line 2: The same translation files are inelegantly copied from the frontend until we can move them to a shared module. The global e2e tests check, however, that the files match.

backend/vaa-strapi/src/functions/utils/appSettings.ts
```diff
@@ -0,0 +1,73 @@
+import {type DynamicSettings, dynamicSettings} from 'vaa-shared';
+import {API} from './api';
+
+/**
+ * Gets `results.cardContents` from `dynamicSettings.ts` and returns them in format used in Strapi
+ */
+export function getCardContentsFromFile() {
```
kaljarv, line 7: The `cardContents` settings are a bit involved, so some conversion is needed between the frontend type and the Strapi components.

frontend/src/lib/i18n/init.ts
```diff
@@ -118,22 +115,6 @@ export {defaultLocale};
 // 4. Utility function exports
 /////////////////////////////////////////////////////
 
-const dynamicTranslationsAdded: Record<string, boolean> = {};
```
kaljarv, line 121: We use flat key-text records for overriding the translations, so these can be added using the built-in `addTranslations` method.

frontend/src/lib/i18n/utils/translate.ts
```diff
@@ -17,10 +17,10 @@ export function translate(strings: LocalizedString | undefined | null, locale?:
...
  * @returns The localized content or `undefined`
  */
-export function translateObject<T extends Record<string, unknown> | null | undefined>(
-  obj: T,
-  locale?: string
-) {
+export function translateObject<
+  TObject extends Record<string, unknown> | null | undefined,
+  TValue = TObject extends Record<string, infer V> ? V : never
+>(obj: TObject, locale?: string): TValue | undefined {
```
kaljarv, line 23: The more precise output typing is needed for candidate app FAQ contents.

frontend/src/lib/stores/stores.ts
```diff
@@ -2,25 +2,30 @@ import {derived, get, writable} from 'svelte/store';
...
  */
-export const settings: Readable<AppSettings> = derived(
-  [page],
-  ([$page]) =>
-    ($page?.data?.appSettings
-      ? Object.assign(localSettings, $page.data.appSettings)
-      : localSettings) as AppSettings
+export const settings: Readable<AppSettings> = derived([page], ([$page]) =>
+  $page?.data?.appSettings
+    ? {...dynamicSettings, ...$page.data.appSettings, ...staticSettings}
```
kaljarv, line 20: Static and dynamic settings do not overlap, so the merge order doesn't matter between them.

shared/src/settings/staticSettings.ts
```diff
@@ -0,0 +1,63 @@
...
+      warning: '#e16060',
+      'line-color': '#262626'
+    }
+  },
+  font: {
+    name: 'Inter',
+    url: "https://fonts.googleapis.com/css2?family=Inter:wght@400;700&display=swap')",
+    style: 'sans'
+  },
+  supportedLocales: [
```
kaljarv, line 45: It's a bit unsatisfactory that locales can only be set here, but we need access to the selected locales when initialising `App Customization` in Strapi. See #609

backend/vaa-strapi/src/api/app-setting/content-types/app-setting/schema.json
kaljarv: Yeah, we can't have multi-level hierarchies because components cannot be nested within each other, but I think that's an acceptable limitation.

backend/vaa-strapi/src/functions/utils/appCustomization.ts
kaljarv: Yeah, see #608. Of course, if there's a suitable update to Svelte 4, then we can anticipate this.

PR #579, opened 2024-07-04
Refactor/basic info to follow question answer model

norppaa:
## WHY:

Fixes #562 

## Check off each of the following tasks as they are completed

- [ ] I have reviewed the changes myself in this PR. Please check the [self-review document](https://github.com/OpenVAA/voting-advice-application/blob/main/docs/contributing/self-review.md)
- [ ] I have added or edited unit tests.
- [ ] I have run the unit tests successfully.
- [ ] I have run the e2e tests successfully.
- [ ] I have tested this change on my own device.
- [ ] I have tested this change on other devices (Using Browserstack is recommended).
- [ ] I have tested my changes using the [WAVE extension](https://wave.webaim.org/extension/)
- [ ] I have added documentation where necessary.
- [ ] Is there an existing issue linked to this PR?

**Clean up your git commit history before submitting the pull request!**

PR #580, opened 2024-07-12
refactor: candidate app form inputs

### LINKED ISSUE #319: Refactor FieldGroup component
I'm wondering if there's any added benefit for providing the `fields` property for the `<FieldGroup>` component and iterating over it inside the component instead of just having a default slot into which the relevant fields go? At least in its current formulation of `any` it doesn't seem to offer much benefits.

Also, the rounding of the corners of the `<div>` that contains the field sometimes breaks the focus outline, which is necessary for low-vision accessibility. It would probably be useful to create a `<Field>` component that takes as its properties or slots the label (contents) and the actual input. Then we could apply consistent styling to all fields.

Also `<FieldGroup>` could apply default styling to the `header` and `footer` slots so that they match the ones now defined on the `/candidate/profile` page.

### LINKED ISSUE #557: a11y: fix MultilangTextInput label targeting
The `<laber for={id}>` model creates errors when the localized fields are expanded.

This could be corrected by using `labelledby` in the `<textarea>`s instead of defining the `for` attribute.

kaljarv:
## WHY:

- Fixes #319 
- Fixes #557

Refactors the multitude of Candidate App components currently handling input fields into reusable, generic components.

Currently, only contains a draft in principle of a common `Input` component for handling all field types.

### DIFF
--- a/frontend/src/lib/components/input/Input.type.ts
+++ b/frontend/src/lib/components/input/Input.type.ts
@@ -1,0 +1,109 @@
+import type {SvelteHTMLElements} from 'svelte/elements';
+import type {Photo} from '$lib/types/candidateAttributes';
+
+export type InputProps =
+  | ({
+      type: 'text';
+    } & InputPropsBase<string> &
+      SvelteHTMLElements['input'])
+  | ({
+      type: 'text-multilingual';
+    } & InputPropsBase<LocalizedString> &
+      SvelteHTMLElements['input'])
+  | ({
+      type: 'textarea';
+    } & InputPropsBase<string> &
+      SvelteHTMLElements['textarea'])
+  | ({
+      type: 'textarea-multilingual';
+    } & InputPropsBase<LocalizedString> &
+      SvelteHTMLElements['textarea'])
+  | ({
+      type: 'number';
+    } & InputPropsBase<number> &
+      SvelteHTMLElements['input'])
+  | ({
+      type: 'date';
+    } & InputPropsBase<Date> &
+      SvelteHTMLElements['input'])
+  | ({
+      type: 'boolean';
+    } & InputPropsBase<boolean> &
+      SvelteHTMLElements['input'])
+  | ({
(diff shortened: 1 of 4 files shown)


frontend/src/lib/components/input/Input.type.ts
```diff
@@ -0,0 +1,109 @@
...
+   */
+  id?: string;
+  /**
+   * Additional info displayed below the input.
+   */
+  info?: string;
+  /**
+   * If `locked` the input will be disabled and a lock icon is displayed.
+   */
+  locked?: boolean;
```
anolpe, line 81: Instead of having `locked` and `disabled` separately maybe there could be something like `state: 'enabled' | 'disabled' | 'locked'` because I don't know if it makes sense that e.g. `disabled` can be changed while `locked = true`.

kaljarv: True, it's a bit superfluous, but it was included due to `disabled` being available by default as part of `restProps` forwarding to `input` elements. There might be a use case, though, for the field being temporarily `disabled` and permanently `locked` with the latter being accompanied with the icon as well. Otherwise, we could lose the whole `locked` prop and just use `disabled`.

kaljarv:
Consider:

- Make `Input` accept only a `Question` and `Answer` as inputs
- Make different subcomponents for different input types which in turn accept more detailed properties, such as `options` for a select
- Enable `localStorage` saving for all types and handle this in `Input`

norppaa:
Also, we should consider the following things:
- define and export getValue() function for input components so that the consumer could get the current value of a component on demand and that way not use value bindings .
- Use type Date for the date input component
- Use same typing for photo input component as well
See #579

kaljarv:
> Sorry, added this originally to the wrong PR 😅 

One layout bug that could be fixed at the same time, is the hide/show password icon overlapping the contents:

image

The lock icon, on the other hand, works as it should:

image

Issue #507, opened 2024-05-17
build: add an automatic check for missing translations when merging into main


kaljarv:
Check that all of the translation keys in all locales (and fields in `App Labels`) match, before merging into `main`.

kaljarv:
New thing to consider:
- Add a component to App Labels in Strapi that can be used to overwrite any of the translations which are not included there by default. Maybe a multilple component of the form:

`Translation key` + `Translated string`

A question is how should we define the target language: as a part of the key, as its own field, or have the translated string be a `LocalizedString` json object?

kaljarv:
Remove all localisation stuff from Strapi so as not to mix things up.

PR #581, opened 2024-07-12
feat: add generated TranslationKey type

anolpe:
## WHY:

Advances #507

TranslationKey type has been added which enables code completion with `$t(key: TranslationKey)`. It can be generated with `yarn generateTranslationKeyType`. Check has been added to CI to make sure that the type is up to date.


frontend/tools/translationKey/generateTranslationKeyType.ts
```diff
@@ -0,0 +1,52 @@
+import fs from 'fs';
+import path from 'path';
+
+/**
+ * Recursive function which returns sorted array of flattened keys.
+ *
+ * For example `{a: 'abc', b: {c: 'def'}}` is returned as `['<prefix>.a', '<prefix>.b.c']`
+ */
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+const flattenKeys = (obj: any, prefix: string): Array<string> => {
```
kaljarv, line 10: Sorry, I didn't mention this before, but let's use the `function flattenKeys(...)` format for functions instead of `const ...`, because it makes the code more legible and to my knowledge there are no real benefits anymore of using the latter format.

This way we can also move the functions to the bottom of the file for added legibility.

frontend/tools/translationKey/generateTranslationKeyType.ts
```diff
@@ -0,0 +1,52 @@
+import fs from 'fs';
+import path from 'path';
+
+/**
+ * Recursive function which returns sorted array of flattened keys.
+ *
+ * For example `{a: 'abc', b: {c: 'def'}}` is returned as `['<prefix>.a', '<prefix>.b.c']`
+ */
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
```
kaljarv, line 9: Maybe we could not use `any` but instead this:
```
interface Translations {
  [key: string]: Translations | string
} 
```
Cannot be a type, though, because of circular referencing.

frontend/tools/translationKey/generateTranslationKeyType.ts
```diff
@@ -0,0 +1,52 @@
...
+
+// Array of flattened translation keys
+const translationFilesKeys = filenames.flatMap((filename) => {
+  return getFlattenedTranslationKeys(locales[0], filename);
+});
+
+const langKeys = locales.map((locale) => `lang.${locale}`);
+const translationKeys = [...translationFilesKeys, ...langKeys].sort();
+
+const output = `export type TranslationKey =\n  | ${translationKeys.map((s) => `'${s}'`).join('\n  | ')};\n`;
```
kaljarv, line 50: We could add a comment to the start of the file like:

```
/* Auto-generated by `/frontend/tools/translationKey/generateTranslationKeyType.ts` */
```

frontend/tools/translationKey/generateTranslationKeyType.ts
```diff
@@ -0,0 +1,52 @@
...
+
+fs.writeFileSync(path.join('src', 'lib', 'types', 'translationKey.ts'), output);
```
kaljarv, line 52: Maybe move the actual path to the top of the file as a `const`. We could also put it in a `generated` subfolder within `types` (but use an `index.ts` to export it from `types` so that the import would be `import type { TranslationKey } from '$types'`), as there might be other auto-generated types in the future as well.

.github/workflows/main.yaml
```diff
@@ -48,6 +48,9 @@ jobs:
       - name: "Run ESlint check on frontend"
         run: yarn lint
 
+      - name: "Check that TranslationKey type is up to date"
```
kaljarv, line 51: Maybe 'Generate the TranslationKey type from translation json files' because this is not a check per se.

kaljarv (changes requested):
Just a few, mainly code style thingies. Otherwise looks good!

anolpe:
I will rebase the `tmp` commit to the first commit before merging.

kaljarv (approved):
LGTM!

Issue #340, opened 2024-01-26
Building the app with Docker creates a lot of volumes


johannalaitila:
There are a lot Docker volumes being created when building the app. Investigate why this is and if it could be reduced.

johannalaitila:
@TrixterTheTux if I remember correctly, were you thinking that potentially node modules are being copied to a new volume every time the app is built?

johannalaitila:
Note for self - might be related to this:
<img width="880" alt="image" src="https://github.com/OpenVAA/voting-advice-application/assets/14262630/e2e4f913-a1a1-4b49-ab6e-e205583c2b67">
https://github.com/docker/compose/issues/1487

TrixterTheTux:
I suspect the cause is https://github.com/OpenVAA/voting-advice-application/blob/main/backend/vaa-strapi/docker-compose.dev.yml#L34 (with similar stuff inside the frontend compose), so what happens is that we create an unnamed volume each time the environment gets setup (due to my previous changes of --force-recreate, which happens to increase the frequency of this becoming an issue).

Issue #530, opened 2024-05-30
build: monorepo structure and submodules


kaljarv:
Check the current repo and suggest possible changes for:

- [ ] Separate pseudo-modules (`/frontend/src/lib/voter/vaa-matching`, `/frontend/src/lib/voter/vaa-filters` and future `vaa-data`) into packages
- [ ] Backend and frontend separation
- [ ] Common folder shared between front- and backend, currently there's a symlinked `/shared` folder
- [ ] `/backend/vaa-strapi/strapi-plugin-import-export-entries`

PR #596, opened 2024-08-07
Use yarn workspaces and make shared directory a module.

gogarufi:
## WHY:

Fixes [Issue link](https://github.com/OpenVAA/voting-advice-application/issues/530#issue-2324842837)

Yarn workspaces are now used within the project:

[README](https://github.com/OpenVAA/voting-advice-application/compare/feature/make-shared-npm-module?expand=1#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5R43)

Shared directory now is a separate module/workspace `vaa-shared`:

[README](https://github.com/OpenVAA/voting-advice-application/compare/feature/make-shared-npm-module?expand=1#diff-8e14ddd6c044992ed17f785e7e9eb16ae030e240c11f026a76705af0dcc5831bR5)


.github/workflows/main.yaml
```diff
@@ -36,11 +36,16 @@ jobs:
         uses: actions/setup-node@v3
         with:
           node-version: 18.x
-          cache: 'yarn'
-          cache-dependency-path: ./frontend/yarn.lock
```
gogarufi, line 40: No longer necessary as cache action will pick up the shared `yarn.lock` from the root by default.

.github/workflows/main.yaml
```diff
@@ -36,11 +36,16 @@ jobs:
...
+          cache: "yarn"
+
+      - name: "Install shared module dependencies"
+        run: yarn workspace vaa-shared install --frozen-lockfile
+
+      - name: "Build shared module"
+        run: yarn workspace vaa-shared build
 
       - name: "Install Frontend dependencies"
-        run: yarn install
```
gogarufi, line 43: Good practice, also helps to avoid unwanted cross-OS related (Mac - Linux) changes within the yarn.lock file.

.github/workflows/main.yaml
```diff
@@ -36,11 +36,16 @@ jobs:
...
+        run: yarn install --frozen-lockfile
```
gogarufi, line 48: That's just a good practice, also helps to avoid unwanted cross-OS changes within the lock file.

backend/vaa-strapi/docker-compose.dev.yml
```diff
@@ -15,7 +15,7 @@ services:
       API_TOKEN_SALT: ${API_TOKEN_SALT}
       ADMIN_JWT_SECRET: ${ADMIN_JWT_SECRET}
       JWT_SECRET: ${JWT_SECRET}
-      DATABASE_HOST: ${DATABASE_HOST}
+      DATABASE_HOST: postgres
```
gogarufi, line 18: The reason being that within the compose set the DB host matches the PostgresDB service name and always stays the same.

backend/vaa-strapi/tsconfig.json
```diff
@@ -1,6 +1,7 @@
 {
   "extends": "@strapi/typescript-utils/tsconfigs/server",
   "compilerOptions": {
+    "moduleResolution": "Node10",
```
gogarufi, line 4: The 'Node' resolution has been deprecated and replaced by 'Node10'. Just a nitpick.

shared/tsconfig.tsbuildinfo
jaguarfi, line 1: just for learning purposes.. What is this file? 😅

shared/README.md
```diff
@@ -1,3 +1,21 @@
...
+yarn build
+```
+
+...or from the project's root directory:
+
+```bash
+yarn workspace vaa-shared build
+```
+
+Currently ESM version is used by `vaa-frontend` and CommonJS by `backend/vaa-strapi` modules.
```
jaguarfi, line 19: Out of curiosity 😅 . Why does the backend/vaa-strapi project use CommonJS?

README.md
```diff
@@ -40,6 +40,52 @@ You can run the whole application in a single Docker image, or run the frontend
 - See [the frontend Readme](frontend/README.md) for instructions on running the frontend individually.
 - See [the backend Readme](/backend/vaa-strapi/README.md) for instructions on running the backend individually.
 
+The project consists of several yarn workspaces (each is a separate module):
+
+- vaa-shared
+- vaa-strapi
+- vaa-frontend
+
+All workspaces share a single `yarn.lock` file located at the project root.
```
jaguarfi, line 49: Nice 👍🏽

README.md
```diff
@@ -40,6 +40,51 @@ You can run the whole application in a single Docker image, or run the frontend
...
+... and then:
+
+```bash
+yarn playwright install
+yarn test
+```
+
+If you encounter any unexpected issues with the E2E tests, make sure to bring down the Docker stack properly to reseed the DB with the original mock data (more on mock data [here](./backend/vaa-strapi/README.md#mock-data)).
+
+To bring down the Docker stack properly (delete all containers, images and named volumes which include backend DB volume with potentially seeded mock data) run:
```
gogarufi, line 82: This was causing an issue for me where even after dropping related containers mock data would not be reseeded, because named Postgres volume still persisted with the mock data in it and was reused by the corresponding container.

backend/vaa-strapi/.env.example
```diff
@@ -3,19 +3,19 @@
 ## Use the .env file in the project root when using Docker
 
 # Backend dependencies
-STRAPI_HOST=0.0.0.0
+STRAPI_HOST=localhost
```
gogarufi, line 6: This needs to be `localhost` in order for the Docker service to be accessible from outside its own Docker compose stack (which is exactly what we need if we spin up only the backend service with this local `.env` file).

backend/vaa-strapi/.env.example
```diff
@@ -3,19 +3,19 @@
...
 STRAPI_PORT=1337
 APP_KEYS="toBeModified1,toBeModified2"
 API_TOKEN_SALT=tobemodified
 ADMIN_JWT_SECRET=tobemodified
-JWT_SECRET=tobemodified
-DATABASE_HOST=postgres
+JWT_SECRET=tobemodified2
+DATABASE_HOST=localhost
```
gogarufi, line 12: This needs to be `localhost` in order for the Docker service to be able to access a DB service which is run outside of its own Docker compose stack (which is exactly what we need if we spin up only the backend service with this local `.env` file).

docker-compose.dev.yml
```diff
@@ -19,7 +19,7 @@ services:
       API_TOKEN_SALT: ${API_TOKEN_SALT}
       ADMIN_JWT_SECRET: ${ADMIN_JWT_SECRET}
       JWT_SECRET: ${JWT_SECRET}
-      DATABASE_HOST: ${DATABASE_HOST}
+      DATABASE_HOST: postgres
```
gogarufi, line 22: The reason being that within the compose set the DB host matches the PostgresDB service name `postgres`, so it shall not be defined by env variables.

frontend/.env.example
```diff
@@ -2,5 +2,5 @@
 ## unless you're running the front separately outside of Docker.
 ## Otherwise please use the .env file in project root.
 
-VITE_BACKEND_URL=
-VITE_PUBLIC_BACKEND_URL=
+VITE_BACKEND_URL=http://localhost:1337
```
gogarufi, line 5: We need these to run frontend as a separate service locally.

frontend/tsconfig.json
```diff
@@ -10,7 +10,7 @@
     "sourceMap": true,
     "strict": true,
     "preserveSymlinks": true,
-    "types": ["vitest/globals"],
-    "moduleResolution": "Node"
```
gogarufi, line 14: Svelte recommends to stick with the more future-proof 'Bundler' resolution mode that it has defined by default.

package.json
```diff
@@ -1,8 +1,9 @@
 {
+  "private": true,
   "scripts": {
     "dev": "docker compose -f docker-compose.dev.yml up -d --force-recreate --build --wait",
     "dev:attach": "docker compose -f docker-compose.dev.yml up --force-recreate --build",
-    "dev:down": "docker compose -f docker-compose.dev.yml down --remove-orphans",
+    "dev:down": "docker compose -f docker-compose.dev.yml down -v --rmi all --remove-orphans",
```
gogarufi, line 6: This command was leaving behind named volumes and images.

shared/README.md
gogarufi: Good question! We use this `strapi-plugin-import-export-entries` plugin, which is a CommonJS module. So we are kinda stack with having to support CommonJS over ESM. 

Internally Strapi devs chose to stick with CommonJS for reasons unknown, they replied in a GitHub thread that they decided it is not worth switching (years ago). :(

package.json
kaljarv: Maybe this resolves #340 as well!

frontend/Dockerfile
```diff
@@ -1,35 +1,52 @@
...
-COPY --from=build ./usr/src/app/yarn.lock .
-COPY --from=build ./usr/src/app/build .
+WORKDIR /opt
+
+COPY --from=build ./opt/frontend/package.json ./frontend
+COPY --from=build ./opt/yarn.lock .
+COPY --from=build ./opt/frontend/build .
 
-RUN yarn install --production
+RUN yarn workspace vaa-frontend install --production --frozen-lockfile
```
kaljarv, line 49: Strangely enough, this command fails when deploying to Render, but not others involving workspaces.

Logs filtered by "workspace":

Screenshot 2024-08-09 at 19 06 48

The whole command 23:

Screenshot 2024-08-09 at 19 07 07

kaljarv (commented):
Looks very good! I only caught a typo, but when deploying on Render there was an error. I'll get you creds to log in there so you can check the logs more thoroughly.

package.json
gogarufi: Yes it does. I mean, if the command is used every time. In case someone drops container stacks manually but leaves images or volumes - they will stay there until deleted manually as well. Otherwise, the command will clean everything that docker compose creates. :)

docs/docker-setup-guide.md
```diff
@@ -6,28 +6,39 @@ backend into a single image that can be built with just a few commands.
...
+```
 
-Finally, run `yarn install` and `yarn dev` in the project root folder.
-These commands will create a Docker image that has the frontend, the backend and a database in a single bundle.
+Then install dependencies for all workspaces:
+
+```bash
+yarn install
+```
+
```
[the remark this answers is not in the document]
gogarufi: I agree on the hot-reloading, but I am not sure about the purpose of this docker setup guide otherwise. I mean does this document attempt to explain how to start the services using existing docker compose yaml files or does this document attempt to describe/explain what steps Dockerfiles contain within themselves? 

I would say that Dockerfiles should explain themselves (they are readable enough and if something is too complex we could leave a comment in there). So this document should explain how to start the services using docker compose files. In that case - we don't need to go into detail of what happens inside the Dockerfiles (which modules are being build and which dependencies are being installed), unless of course one needs to do something manually prior to docker compose execution. 

What do you think?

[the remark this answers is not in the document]
kaljarv: Strictly speaking it's true that the note is not about Docker setup per se, but as this doc de facto serves as a "getting started with development" guide, it might be useful to mention that bit here as well. When reorganising the documentation later, we could separate the two.

(Just to make sure what I mean by IDE availaibility is that VS Code is able to link to the module. I guess another option would to just bundle that command into the `dev` script…)

Anyway, if you feel like adding this isn't kosher, we can leave it out and maybe add an item in troubleshooting?

frontend/Dockerfile
kaljarv: Great, this works now!

kaljarv (approved):
LGTM! You can choose yourself how to deal with the docs question.

docs/docker-setup-guide.md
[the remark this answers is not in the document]
gogarufi: Right, I see your point on the IDE. I feel like these general notes belong to README at the root of the project. Which should explain the structure of the monorepo, dev tools setup and gotchas. I will try and do that now.

For the future, I would say that this document shall go in more detail on Docker specific things, f.e. that each module has it's own .env, compose yaml and Docker files. Explaining step by step how they can be started locally separately or altogether from the project's root with compose yaml. 🤔 

I would gladly implement my own vision if you give me the green light - I enjoy writing in general, but it could take me some time, not sure it has the priority now?

shared/tsconfig.tsbuildinfo
gogarufi: I wouldn't be able to explain better myself, but in general just gives a small speed boost for `tsc` transpiler or a big one if the project is larger. I would say just a good practice at this point to have it.

https://www.typescriptlang.org/tsconfig/#tsBuildInfoFile

docs/docker-setup-guide.md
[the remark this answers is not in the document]
kaljarv: Great, that sounds logical!

If you feel like, it'd be easiest to write the Docker docs now that everything's still fresh in the memory, go ahead and do it. Otherwise, let's move that to a separate issue for later.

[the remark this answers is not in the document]
gogarufi: Alright, I updated the `README` in the root to describe the different module resolution mechanisms that take place within the monorepo. I will create a separate issue on updating the Docker docs. :)

shared/tsconfig.tsbuildinfo
jaguarfi: Thanks. I will read more about that specific file.

PR #599, opened 2024-08-12
Chore/update linter and prettier to match styleguide

norppaa:
## WHY:

Updates eslint and prettier to match the new [ts styleguide](https://mkosir.github.io/typescript-style-guide/)


.prettierignore
```diff
@@ -0,0 +1,25 @@
...
+docs/
+node_modules/
+playwright/
+playwright-report/
+test-results/
+shared/build/
+shared/node_modules/
+
+README.md
+dockerc-compose.*
```
kaljarv, line 12: => docker-compose.*

.prettierignore
```diff
@@ -0,0 +1,25 @@
+frontend/
```
kaljarv, line 1: I guess these are here so that if we run prettier from the root, the submodules will not be affected?

frontend/src/lib/utils/breakpoints.ts
```diff
@@ -2,10 +2,12 @@
...
  * NB. Currently, the values used here are hardcoded and should be manually kept in line width the Tailwind definitions.
  */
-export enum Breakpoints {
-  sm = 640,
-  md = 768,
-  lg = 1024,
-  xl = 1280,
-  '2xl' = 1536
-}
+export const BREAKPOINTS: Record<'sm' | 'md' | 'lg' | 'xl' | '2xl', number> = {
```
kaljarv, line 5: Do we need the `Record...` typing here?

frontend/src/lib/utils/navigation/route.ts
```diff
@@ -2,37 +2,73 @@
...
-  CandAppPreview = 'candidate/preview',
-  CandAppProfile = 'candidate/profile',
-  CandAppQuestions = 'candidate/questions',
-  CandAppQuestionEdit = 'candidate/questions/edit',
-  CandAppRegister = 'candidate/register',
-  CandAppResetPassword = 'candidate/password-reset',
-  CandAppSettings = 'candidate/settings',
-  Candidate = 'candidates',
-  Candidates = 'candidates',
+export type RouteOptions =
```
kaljarv, line 5: Likewise, I don't we need this type (or need to apply it to the `const`).

frontend/src/lib/utils/navigation/route.ts
```diff
@@ -2,37 +2,73 @@
...
+  | 'Question'
+  | 'QuestionCategory'
+  | 'Questions'
+  | 'ResultCandidate'
+  | 'ResultParty'
+  | 'Results'
+  | 'Statistics'
+  | '_Test';
+
+export const Route: Record<RouteOptions, string> = {
```
kaljarv, line 39: Could you use `CAMEL_CASE` for all `consts` while reserving `CamelCase` for the associated types? I.e. this would become `const ROUTE` and the type below just `type Route`.

frontend/src/lib/utils/navigation/route.ts
```diff
@@ -2,37 +2,73 @@
...
+  Privacy: 'privacy',
+  Question: 'questions',
+  QuestionCategory: 'questions/category',
+  Questions: 'questions',
+  ResultCandidate: 'results/candidate',
+  ResultParty: 'results/party',
+  Results: 'results',
+  Statistics: 'results/statistics',
+  _Test: '_test'
+};
```
kaljarv, line 72: Add `as const`.

frontend/src/lib/voter/vaa-filters/src/group/combineResults.ts
```diff
@@ -3,12 +3,14 @@ import type { MaybeWrapped } from '../entity';
 /**
  * A logic operator for combining results from multiple filters.
  */
-export enum LogicOp {
+export const LogicOp: Record<'And' | 'Or', number> = {
```
kaljarv, line 6: Ditto: camel case and typing can likely be removed.

kaljarv (changes requested):
Looks good overall! Just a couple of small things and a question.

.prettierignore
norppaa: Yes. [Combining the  config files](https://github.com/eslint/eslint/discussions/16960) is also an option, but I don't think I have time for that any longer.

norppaa: Yes. The config files could probably be combined if we wanted to and then there would be no need to specify this, but then some of the svelte dependencies would have to be downloaded to the root, and that is probably not something we want.

frontend/src/lib/voter/vaa-filters/src/group/combineResults.ts
norppaa: Typing is neede

kaljarv:
At the same time, also fork the [TS style guide](https://mkosir.github.io/typescript-style-guide/) with the changes agreed in #624

kaljarv (approved):

.editorconfig
kaljarv, line 1: Moved this to repo root, where it'll also be parsed by prettier which is now run only from the root of the repo.

.gitignore
```diff
@@ -2,9 +2,4 @@
 .DS_STORE
 .env
 !.env.example
-node_modules
```
kaljarv, line 5: The tests are moved to their own folder (and workspace).

.husky/pre-commit
```diff
@@ -1,6 +1,7 @@
+yarn buildDeps
```
kaljarv, line 1: We need to build the app-shared module for generateTranslationKeyType to work.

.lintstagedrc.json
```diff
@@ -0,0 +1,8 @@
+{
+  "*.{html,js,jsx,cjs,mjssvelte,ts,tsx,cts,mts,xml,yaml,yml}": [
+    "yarn buildDeps",
+    "prettier --write",
+    "eslint --fix --flag unstable_config_lookup_from_file"
```
kaljarv, line 5: The flag allows us to use the new flat-style eslint config files in subdirectories the same way as the deprecated .eslintrc format. This will become the default behaviour with eslint 10.

.prettierignore
kaljarv, line 1: Now that we handle formatting from the root, we need to collect all prettierignore rules here, because it doesn't allow setting these in subdirectories, uh!

backend/vaa-strapi/.eslintignore
kaljarv, line 1: The depreacted ignores and rc below are combined in the new esling.config.mjs.

backend/vaa-strapi/eslint.config.mjs
```diff
@@ -0,0 +1,34 @@
+import { default as sharedConfig } from '@openvaa/shared-config/eslint';
```
kaljarv, line 1: We're now importing the base settings from @openvaa/shared-config to all the subconfigurations.

backend/vaa-strapi/package.json
```diff
@@ -8,20 +8,11 @@
     "start": "strapi start",
     "build": "strapi build",
     "strapi": "strapi",
-    "lint": "eslint .",
```
kaljarv, line 11: Anything to do with linting or formatting is removed from all of the workspaces and only handled by the scripts in the root package.json.

backend/vaa-strapi/package.json
```diff
@@ -49,15 +39,13 @@
   },
   "license": "MIT",
   "devDependencies": {
-    "@babel/eslint-parser": "^7.24.7",
     "@babel/plugin-transform-modules-commonjs": "^7.24.7",
     "@faker-js/faker": "^8.4.1",
-    "@typescript-eslint/parser": "^7.14.1",
-    "eslint": "^8.57.0",
-    "eslint-config-prettier": "^8.10.0",
+    "@openvaa/shared-config": "^1.0.0",
```
kaljarv, line 44: @openvaa/shared-config is now needed as a devDep in all workspaces.

backend/vaa-strapi/src/api/answer/routes/answer.ts
```diff
@@ -3,53 +3,44 @@
...
 
 import { factories } from '@strapi/strapi';
-import { restrictResourceOwnedByCandidate, restrictPopulate, restrictFilters, electionCanEditAnswers } from '../../../util/acl';
+import {
+  electionCanEditAnswers,
+  restrictFilters,
+  restrictPopulate,
+  restrictResourceOwnedByCandidate
+} from '../../../util/acl';
+import { Generic } from '../../../util/acl.type';
```
kaljarv, line 12: Incomplete typing for the ACL functions is also fixed. This `Generic` type is something of an ugly workaround Strapi's incomplete typing. See more in the comment for the `util/acl` file.

backend/vaa-strapi/src/api/candidate/routes/candidate.ts
```diff
@@ -1,46 +1,39 @@
...
+        restrictFilters([])
+      ]
     },
     update: {
       policies: [
         // Allow only updating candidate itself
-        async (ctx: any, config, {strapi}) => {
-          const {id} = ctx.params;
-
+        async ({ params, state }: StrapiContext, config, { strapi }) => {
```
kaljarv, line 30: Better typing and null guards for this ACL method.

backend/vaa-strapi/src/util/acl.type.ts
```diff
@@ -0,0 +1,46 @@
+import type { factories } from '@strapi/strapi';
+import type { Strapi } from '@strapi/types';
+
+// Strapi 5: Hopefully, these types can be removed
```
kaljarv, line 4: These types add a bit of rigour to the ACL functions but hopefully we can get rid of these when updating to Strapi 5, which should also allow us to use the types for the Content Types.

backend/vaa-strapi/tsconfig.json
```diff
@@ -1,10 +1,9 @@
 {
   "extends": "@strapi/typescript-utils/tsconfigs/server",
   "compilerOptions": {
-    "moduleResolution": "Node10",
```
kaljarv, line 4: Removing this bc the deprecated modRes is defined in @strapi/typescript-utils/tsconfigs/server. When we update Strapi, we don't want this override to remain.

frontend/vitest.config.ts
kaljarv, line 1: Even though we've removed vitest from the modules’ devDeps, we can still define vitest workspaces by just adding a vitest.config.ts in any folder tree.

package.json
```diff
@@ -1,28 +1,36 @@
 {
   "private": true,
   "scripts": {
+    "buildDeps": "yarn workspace @openvaa/app-shared build && true",
```
kaljarv, line 4: Some commands require the app-shared module to be built to run properly.

The && true in the end prevents command line arguments in the pre-commit hook from being passed to this command.

package.json
```diff
@@ -1,28 +1,36 @@
...
     "dev:attach": "docker compose -f docker-compose.dev.yml up --force-recreate --build",
     "dev:down": "docker compose -f docker-compose.dev.yml down -v --rmi all --remove-orphans",
     "dev:stop": "docker compose -f docker-compose.dev.yml stop",
     "prepare": "husky",
     "prod": "docker compose -f docker-compose.yml -p vaa-prod up -d --build",
-    "test": "playwright test"
+    "test": "yarn workspace @openvaa/tests test",
+    "format": "yarn buildDeps && prettier --write .",
+    "format:check": "yarn buildDeps && prettier --check .",
+    "lint": "yarn buildDeps && eslint --flag unstable_config_lookup_from_file backend/vaa-strapi/src backend/vaa-strapi/tests frontend packages tests",
```
kaljarv, line 14: The flag's needed until eslint 10 to use nested eslint.configs. We need to enumerate the folders in the backend/vaa-strapi we want to lint because, some of the other folders contain eslint.configs that cause an error with eslint 9 – even if we mark them as ignored!

packages/core/package.json
kaljarv, line 1: The shared modules now have a minimal structure with only the build script in them and few devDeps. Even vitest is a bit optional, because we run the tests from the root but it's needed to prevent linter errors for the imports from vitest in the test files within the package.

These could maybe made globally available, though.

package.json
```diff
@@ -1,28 +1,39 @@
 {
   "private": true,
   "scripts": {
+    "build:app-shared": "yarn workspace @openvaa/app-shared build && true",
```
kaljarv, line 4: Adds scripts for building shared dependencies which are needed for linting or generating the translation key types.

package.json
```diff
@@ -1,28 +1,39 @@
...
     "dev:stop": "docker compose -f docker-compose.dev.yml stop",
     "prepare": "husky",
     "prod": "docker compose -f docker-compose.yml -p vaa-prod up -d --build",
-    "test": "playwright test"
+    "test:e2e": "yarn workspace @openvaa/tests test",
+    "test:unit": "vitest run",
+    "test:unit:watch": "vitest",
+    "format": "yarn build:app-shared && prettier --write .",
+    "format:check": "yarn build:app-shared && prettier --check .",
+    "lint:check": "yarn build:app-shared && prettier --write . && eslint --flag unstable_config_lookup_from_file --fix backend/vaa-strapi/src backend/vaa-strapi/tests frontend packages tests",
```
kaljarv, line 17: In addition to the flag, we also need to painfully enumerate the folders within backend that we want to check because some of the strapi-generated folder contain eslint.configs which will cause an error – even if we mark these folders as ignored in the eslint.config in /backend/vaa-strapi! With eslint 10 we can most likely change this to just `eslint .`.

frontend/package.json
```diff
@@ -9,65 +9,59 @@
...
-    "algorithm-example": "ts-node-esm --experimentalSpecifierResolution node src/matching/examples/example.ts",
-    "generateTranslationKeyType": "node --no-warnings=ExperimentalWarning --loader ts-node/esm tools/translationKey/generateTranslationKeyType.ts && prettier --write src/lib/types/generated/translationKey.ts"
-  },
-  "lint-staged": {
-    "**/*.{svelte,css,html,js,json,md,ts,xml,yaml,yml}": [
-      "prettier --write",
-      "eslint"
-    ]
+    "algorithm-example": "tsx src/matching/examples/example.ts",
+    "generate:translation-key-type": "tsx tools/translationKey/generateTranslationKeyType.ts && prettier --write src/lib/types/generated/translationKey.ts"
```
kaljarv, line 13: Rename yarn scripts to kebab-case and harmonizes (though not completely) usage of colons in script names to conform to the behaviour or coming versions of yarn where any script with a colon can be run globally from within any workspace.

packages/shared-config/package.json
kaljarv, line 1: This package simplifies a bit having uniform configs for eslint, prettier and typescript across the modules. These configs can be used like `"extends": "@openvaa/shared-config/ts"` or a similar way.

tests/package.json
kaljarv, line 1: E2e (and some other global) tests are now collected in their own package with it's own dependencies. The root packages has a  `test:e2e` script which will invoke the `test` script here.

vitest.workspace.ts
kaljarv, line 1: In a manner suitable for a monorepo, tests are now separated into workspaces (by just adding a config file next to package.json). To run all of them use `yarn test:unit` in the root and see the results nicely tagged with the module names.

PR #600, opened 2024-08-14
feat: Vertical layout for candidate text answers

gogarufi:
## WHY:

Fixes [Issue link](https://github.com/orgs/OpenVAA/projects/7?pane=issue&itemId=69315433)

For answers to questions with "text" type:

- we use vertical layout (as in "no columns but rows") now.
- label has unlimited width to compliment / benefit from the vertical layout.
- the vertical layout item got extra bottom padding to separate itself from the following horizontal items visually.

I also increased the column width for all labels - as of now the answers column is supposed to contain only short text and the labels can benefit from some extra width for readability.

Screenshot 2024-08-14 at 17 00 49

### DIFF
--- a/frontend/src/lib/components/entityDetails/InfoItem.svelte
+++ b/frontend/src/lib/components/entityDetails/InfoItem.svelte
@@ -1,6 +1,8 @@
 <script lang="ts">
   /** The info label */
   export let label: string;
+  /** The info layout mode */
+  export let vertical = false;
 </script>
 
 <!--
@@ -10,6 +12,7 @@ Used to show a label-content pair in a Candidate's basic information.
 ### Properties
 
 - `label`: the label of the information
+- `vertical`: layout mode for the item
 
 ### Slots
 
@@ -24,10 +27,25 @@ Used to show a label-content pair in a Candidate's basic information.
 ```
 -->
 
-<div class="infoItem grid grid-flow-col grid-cols-[min-content_auto] justify-start gap-md">
+<div class="grid justify-start gap-md {vertical ? 'vertical-grid' : 'horizontal-grid'}">
   <!-- pt-[0.3rem] matches the baselines of the small label and the content text -->
-  <div class="small-label w-min min-w-[7rem] pt-[0.3rem] text-left align-top">{label}</div>
+  <div
+    class="test-label small-label pt-[0.3rem] text-left align-top {vertical
+      ? 'w-max'
+      : 'min-w-[10rem]'}">
+    {label}
+  </div>
   <div class="overflow-hidden align-top">
     <slot />
   </div>
 </div>
+
+<style lang="postcss">
+  .vertical-grid {
+    @apply grid-flow-row grid-rows-[min-content_auto] pb-8;
+  }
+
+  .horizontal-grid {
+    @apply grid-flow-col grid-cols-[min-content_auto];
+  }
+</style>
--- a/frontend/src/lib/components/entityDetails/EntityInfo.svelte
+++ b/frontend/src/lib/components/entityDetails/EntityInfo.svelte
@@ -70,7 +70,7 @@ Used to show an entity's basic info in an `EntityDetails` component.
       {#if nonLinkQuestions.length}
         {#each nonLinkQuestions as question}
           {#if $settings.entityDetails.showMissingAnswers[entityType] || getAnswer(entity, question) != null}
-            <InfoItem label={question.text}>
+            <InfoItem label={question.text} vertical={question.type === 'text'}>
               <InfoAnswer {entity} {question} />
             </InfoItem>
           {/if}


frontend/src/lib/components/entityDetails/InfoItem.svelte
```diff
@@ -24,10 +25,20 @@ Used to show a label-content pair in a Candidate's basic information.
 ```
 -->
 
-<div class="infoItem grid grid-flow-col grid-cols-[min-content_auto] justify-start gap-md">
+<div class="grid justify-start gap-md {vertical ? 'vertical-grid' : 'horizontal-grid'}">
```
gogarufi, line 28: I also tried using [conditional CSS directives from Svelte](https://sveltebyexample.com/class-and-style-directives/) but eventually settled on just a regular conditional (ternary) operator combined with a string insertion.

My reasoning:
- it allow to apply different classes at once (either/or) rather than have to conditional CSS Svelte directives like so:

```
class:vertical-grid={vertical}
class:horizontal-grid={!vertical}
```
- conditional CSS directives are not possible to define when one wants to apply multiple classes based on a single condition, whereas the approach with ternary operator is not limited in this case, so it is more flexible and therefor better for uniformity

frontend/src/lib/components/entityDetails/InfoItem.svelte
```diff
@@ -24,10 +25,20 @@ Used to show a label-content pair in a Candidate's basic information.
...
   <!-- pt-[0.3rem] matches the baselines of the small label and the content text -->
-  <div class="small-label w-min min-w-[7rem] pt-[0.3rem] text-left align-top">{label}</div>
+  <div class="small-label pt-[0.3rem] text-left align-top {vertical ? 'w-max' : 'min-w-[10rem]'}">{label}</div>
   <div class="overflow-hidden align-top">
     <slot />
   </div>
 </div>
+
+<style lang="postcss">
```
gogarufi, line 36: Even though Tailwind [recommends to avoid using @apply](https://tailwindcss.com/docs/reusing-styles#avoiding-premature-abstraction) just to avoid listing multiple classes. I believe when we do so within a Svelte component (in an isolated manner) - it is justified, as maintaining our own CSS classes only per component is not that cognitively demanding.

frontend/src/lib/components/entityDetails/InfoItem.svelte
```diff
@@ -24,10 +25,20 @@ Used to show a label-content pair in a Candidate's basic information.
 ```
 -->
 
-<div class="infoItem grid grid-flow-col grid-cols-[min-content_auto] justify-start gap-md">
+<div class="grid justify-start gap-md {vertical ? 'vertical-grid' : 'horizontal-grid'}">
   <!-- pt-[0.3rem] matches the baselines of the small label and the content text -->
-  <div class="small-label w-min min-w-[7rem] pt-[0.3rem] text-left align-top">{label}</div>
+  <div class="small-label pt-[0.3rem] text-left align-top {vertical ? 'w-max' : 'min-w-[10rem]'}">{label}</div>
```
gogarufi, line 30: Strangely enough I could get something like bellow working:

```
<style>
   .vertical-grid .small-label {
      @apply w-max
   }
   ...
</style>
```

I believe the reason is `.small-label` being already a layer defined in `app.css`. Probably it is not supported for the better. :D

frontend/src/lib/components/entityDetails/InfoItem.svelte
```diff
@@ -24,10 +25,25 @@ Used to show a label-content pair in a Candidate's basic information.
...
 -->
 
-<div class="infoItem grid grid-flow-col grid-cols-[min-content_auto] justify-start gap-md">
+<div class="grid justify-start gap-md {vertical ? 'vertical-grid' : 'horizontal-grid'}">
   <!-- pt-[0.3rem] matches the baselines of the small label and the content text -->
-  <div class="small-label w-min min-w-[7rem] pt-[0.3rem] text-left align-top">{label}</div>
+  <div
+    class="test-label small-label pt-[0.3rem] text-left align-top {vertical
+      ? 'w-max'
+      : 'min-w-[10rem]'}">
```
gogarufi, line 33: I noticed that on the "big data refactoring" branch this is already set to 10rem! 👍

frontend/src/lib/components/entityDetails/InfoItem.svelte
kaljarv: Yeah, I think it's fine in general to use @apply in situations like this. In this particular one, however, I think it might also be okay to just stash everything within the ternary op so that one could see all styling happening at a glance, but it's nice this way as well.

kaljarv (approved):
Looks good! Just update the docs and merge.

PR #613, opened 2024-09-21
feat: add AWS SES support for nodemailer

gogarufi:
## WHY:

Allows to send emails via AWS SES.

Nodemailer now uses SES provider, which uses AWS SES HTTP API (based on their official documentation it is faster and more reliable than direct SMTP). This provider also supports buffering to meet AWS SES limit on emails per/second via `sendingRate` option. 

Environmental variables for AWS SES are saved in Render `AWS SES HTTP` env group for now.


backend/vaa-strapi/docker-compose.dev.yml
```diff
@@ -25,6 +25,11 @@ services:
       GENERATE_MOCK_DATA_ON_INITIALISE: ${GENERATE_MOCK_DATA_ON_INITIALISE}
       GENERATE_MOCK_DATA_ON_RESTART: ${GENERATE_MOCK_DATA_ON_RESTART}
       LOAD_DATA_ON_INITIALISE_FOLDER: ${LOAD_DATA_ON_INITIALISE_FOLDER}
+      AWS_ACCESS_KEY_ID: ${AWS_ACCESS_KEY_ID}
```
kaljarv, line 28: These will actually not have any effect now, because the `config/development/plugins.ts` forces use of `maildev`. Maybe we could add another `env` setting to control that, e.g. `USE_MAILDEV_IN_DEVELOPMENT`?

docs/candidate-app/candidates.md
```diff
@@ -26,11 +26,10 @@ You can create a new user using the "Create new entry" button. The following fie
...
-The user gets an email with a link to reset their password using the forgot password functionality on the login page. The frontend URL in the emails is configured in `.env` with the PUBLIC_FRONTEND_URL variable, and the SMTP can be configured using the following variables:
-- `SMTP_HOST`: the hostname the SMTP server runs on
-- `SMTP_PORT`: the port of the SMTP server
-- `SMTP_USERNAME`: the username used for authentication
-- `SMTP_PASSWORD`: the password used for authentication
+The user gets an email with a link to reset their password using the forgot password functionality on the login page. The frontend URL in the emails is configured in `.env` with the PUBLIC_FRONTEND_URL variable, and the email service (AWS SES) can be configured using the following variables:
+- `AWS_ACCESS_KEY_ID`: AWS SES user access key
+- `AWS_SECRET_ACCESS_KEY`: AWS SES user secret access key
+- `AWS_REGION`: AWS SES region
 - `MAIL_FROM`: the email address the emails are sent from
```
kaljarv, line 33: We should add a note here (and in `.env.example`) that you need also to set these addresses in Strapi > Settings > Email Templates for all templates, otherwise the functions associated with the templates will fail. Of course, if you can think of an easy way to insert these from `env` that'd be nice, but probably not worth the effort.

kaljarv (approved):
Works like a dream! Just a few notes that could be added.

.env.example
```diff
@@ -15,13 +15,12 @@ DATABASE_SCHEMA=public
...
-SMTP_PORT=587
-SMTP_USERNAME=""
-SMTP_PASSWORD=""
-MAIL_FROM="candidate-app@example.com"
-MAIL_REPLY_TO="candidate-app@example.com"
+# AWS SES Settings
+AWS_ACCESS_KEY_ID="EXAMPLE_KEY"
+AWS_SECRET_ACCESS_KEY="EXAMPLE_SECRET"
+AWS_REGION="eu-north-1"
+MAIL_FROM="no-reply@openvaa.org"
```
kaljarv, line 22: Oh, maybe we can change the examples to:
```
MAIL_FROM="OpenVAA Voting Advice Application <no-reply@openvaa.org>"
MAIL_REPLY_TO="OpenVAA Admin <contact@openvaa.org>"
```

backend/vaa-strapi/docker-compose.dev.yml
kaljarv: Okay! And I let's pass all vars for sure.

kaljarv (approved):
👍

PR #616, opened 2024-10-01
Use AWS S3 as Strapi media storage provider.

gogarufi:
## WHY:

We want to store media files in AWS S3.

- In production we will use AWS S3 as media storage provider for Strapi. The necessary environmental variables are defined in Render in `AWS S3 MEDIA STORAGE` group. 
- In development we will keep using local storage.

**Note:** if one uses AWS S3 to store media while running Strapi locally thumbnails in "Strapi -> Media Library" will no be displayed due to CORS (this is expected behaviour).

The new AWS S3 bucket `static.openvaa.org` is an example (it contains necessary minimal access policies and CORS settings). There is a new CNAME DNS record in CloudFlare to support redirections from URLs like `https://static.open.vaa/public/media/image.png` to the AWS S3 bucket url.

**WIP** Still need to add a few words into the README and in the development document in Google Drive about the whole setup.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -27,6 +27,17 @@ MAIL_REPLY_TO="OpenVAA Admin <contact@openvaa.org>"
 # Maildev settings (used only in dev)
 MAILDEV_PORT=1080
 
+## AWS S3 settings for static content storage
+AWS_S3_ACCESS_KEY_ID="EXAMPLE_KEY"
+AWS_S3_ACCESS_SECRET="EXAMPLE_SECRET"
+AWS_S3_REGION=eu-north-1
+AWS_S3_BUCKET=static-example-com
+
+## Will be used as the base URL for media content uploaded via Strapi's UI.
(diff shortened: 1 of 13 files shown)


kaljarv (changes requested):
:+1: This works nicely otherwise, but we have some functions handling urls with the expectation they're local in the frontend. Could you fix these at the same time?

The relevant files are:
1. [parseImage.ts](frontend/src/lib/api/dataProvider/strapi/utils/parseImage.ts)
2. [PhotoInput.svelte](frontend/src/lib/candidate/components/input/PhotoInput.svelte)

Perhaps, `parseImage` could check if the `URL` provided is relative or absolute and slap in a base `URL` if relative?

The latter is in need of refactoring bc it's now Strapi-dependent, but I guess we can go with a quick fix for now, as it's scheduled for a redo anyway.

kaljarv (approved):
👍

PR #661, opened 2025-01-13
Refactor candidate app layout.

gogarufi:
The candidate app layout was refactored to align with recent voter app changes - with the main idea being to nest and reuse layouts to take advantage of Svelte's approach to re-rendering (or in this case preserving) html while browsing through app pages.

Candidate app now uses the recommended by Svelte cookie-based token auth with two newly added POST actions (API endpoints) on Svelte-server - `login` and `logout`, which set and delete the auth cookie. Svelte candidate context, related pages and layouts, alongside with E2E tests were updated to support the new approach.

Some page components and their properties went away and were merged into corresponding page.svelte or layout.svelte files. Some new routes were introduced as a consequence.

**Drawer related code has been migrated from root `+layout.svelte` into `Layout.svelte` component, since now contexts for candidate and voter apps are separated and bound to their scopes (can not be accessed from outside the apps by the root `layout.svelte` anymore). Hopefully we will be able to move all the shared code from `Layout.svelte` into the root `+layout.svlete` with helps of snippets functionality introduced in the new SvelteKit version - then we can attempt to inject app/context-dependent bits and pieces from within the corresponding app's pages/layouts into the root `+layout.svelte` file.**

TODOs:

- some semi-related minor changes which have been discussed along the way could still be missing


kaljarv (approved):
👍

PR #671, opened 2025-01-28
feat/llm

aleksilassila:
**Definition of Done**
- It can take in a prompt, and return a response, using the OpenAI API
- It uses an extendable object-oriented design so that other AI companies' APIs can be integrated in the future, but only OpenAI will be integrated for now
- It is built as a package in packages/llm, and usage of the package requires to input the API key from the frontend/Sveltekit API
- Unit tests implemented

### DIFF
--- a/packages/llm/src/llm-providers/openai-provider.ts
+++ b/packages/llm/src/llm-providers/openai-provider.ts
@@ -6,16 +6,22 @@ export class OpenAIProvider extends LLMProvider {
   private openai: OpenAI;
   public readonly maxContextTokens: number;
 
-  constructor(options: { model?: string; apiKey: string; maxContextTokens?: number }) {
-    super();
-    this.model = options.model || 'gpt-4o-mini';
-    this.maxContextTokens = options.maxContextTokens || 4096;
-
-    if (!options.apiKey) {
+  constructor({
+    model = 'gpt-4o-mini',
+    apiKey,
+    maxContextTokens = 4096
+  }: {
+    model?: string;
+    apiKey: string;
+    maxContextTokens?: number;
+  }) {
+    if (!apiKey) {
       throw new Error('OpenAI API key is required in constructor options.');
     }
-
-    this.openai = new OpenAI({ apiKey: options.apiKey });
+    super();
+    this.model = model;
+    this.maxContextTokens = maxContextTokens;
+    this.openai = new OpenAI({ apiKey });
   }
 
   async generate({
@@ -77,9 +83,7 @@ export class OpenAIProvider extends LLMProvider {
   }
 
   /**
-   * Estimates the number of tokens in a text string.
-   * Note: This is a simple approximation. For production use,
-   * consider using a proper tokenizer.
+   * Estimates the number of tokens in a text string. Note: This is a simple approximation. For production use, consider using a proper tokenizer.
    */
   async countTokens(text: string) {
     return {
@@ -88,9 +92,7 @@ export class OpenAIProvider extends LLMProvider {
   }
 
   /**
-   * Calculates how many messages can fit within the context window.
-   * Uses a conservative estimate of average tokens per message to ensure
-   * we don't exceed the model's context limit.
+   * Calculates how many messages can fit within the context window. Uses a conservative estimate of average tokens per message to ensure we don't exceed the model's context limit.
    */
   async fitCommentArgsCount(): Promise<number> {
     const averageTokensPerMessage = 50;
@@ -98,35 +100,23 @@ export class OpenAIProvider extends LLMProvider {
   }
 }
 
-function mapToMessageParam(message: { role: string; content: string }): OpenAI.ChatCompletionMessageParam {
-  const role = message.role.toLowerCase();
+function mapToMessageParam({ role, content }: { role: string; content: string }): OpenAI.ChatCompletionMessageParam {
+  const normalizedRole = role.toLowerCase();
 
-  switch (role) {
+  switch (normalizedRole) {
     case 'system':
-      return {
-        role: 'system',
-        content: message.content
-      } as OpenAI.ChatCompletionSystemMessageParam;
+      return { role: 'system', content } as OpenAI.ChatCompletionSystemMessageParam;
 
     case 'user':
-      return {
-        role: 'user',
-        content: message.content
-      } as OpenAI.ChatCompletionUserMessageParam;
+      return { role: 'user', content } as OpenAI.ChatCompletionUserMessageParam;
 
     case 'assistant':
-      return {
-        role: 'assistant',
-        content: message.content
-      } as OpenAI.ChatCompletionAssistantMessageParam;
+      return { role: 'assistant', content } as OpenAI.ChatCompletionAssistantMessageParam;
 
     case 'developer':
-      return {
-        role: 'developer',
-        content: message.content
-      } as OpenAI.ChatCompletionDeveloperMessageParam;
+      return { role: 'developer', content } as OpenAI.ChatCompletionDeveloperMessageParam;
 
     default:
-      throw new Error(`Unsupported role: ${role}`);
+      throw new Error(`Unsupported role: ${normalizedRole}`);
   }
 }


packages/llm/package.json
kaljarv, line 1: Maybe it'd be best for now to copy the build scripts and other definitions from `app-shared`  which is the only other package being built for both esm and cjs so that we only use one paradigm for such.

packages/llm/src/llm-providers/openai-provider.ts
```diff
@@ -0,0 +1,73 @@
...
+  constructor(options: { model?: string; apiKey?: string } = {}) {
+    super();
+    this.model = options.model || 'gpt-4o-mini';
+    this.openai = new OpenAI({
+      apiKey: options.apiKey ?? process.env.OPENAI_API_KEY
+    });
+  }
+
+  // OpenAI models typically have a max context window of around 4096 tokens
+  get maxContextTokens(): number {
```
kaljarv, line 18: Should this be a property that can be set in the constructor options?

packages/llm/src/llm-providers/llm-provider.ts
```diff
@@ -0,0 +1,85 @@
+export enum Role {
+  SYSTEM = 'system',
+  USER = 'user'
+  // ASSISTANT = 'assistant'
+}
+
+export class Message {
```
kaljarv, line 7: It might be best to use named, deconstructable parameters for all class constructors (and other methods) to avoid bugs, such as in `OpenAIProvider`. It's crappy that ts doesn't support that the way python does 😅

packages/llm/tsconfig.base.json
kaljarv, line 1: Could you rename this to just tsconfig.json so it's in line with `app-shared`.

kaljarv (commented):
Looks good thus far!

packages/llm/tsconfig.base.json
edthu: Renamed in 56a4f7f689bb7371fcb16584fdc12c7a6c5ba0e0

packages/llm/package.json
edthu: Refactored llm to style of app-shared in 56a4f7f689bb7371fcb16584fdc12c7a6c5ba0e0

backend/vaa-strapi/tsconfig.json
```diff
@@ -1,6 +1,10 @@
 {
   "extends": "@strapi/typescript-utils/tsconfigs/server",
   "compilerOptions": {
+    "paths": {
```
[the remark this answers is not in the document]
edthu: Removed in 56a4f7f689bb7371fcb16584fdc12c7a6c5ba0e0

backend/vaa-strapi/src/functions/generateMockData.ts
```diff
@@ -668,6 +677,54 @@ async function createCandidateUsers() {
...
+              "title": {
+                "en": "Background"
+              },
+              "visible": true
+            }
+          }
+        }`
+        }
+      ])
+      .then((r) => r);
```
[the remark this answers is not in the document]
edthu: Removed in 56a4f7f689bb7371fcb16584fdc12c7a6c5ba0e0

.env.example
```diff
@@ -56,7 +56,12 @@ GENERATE_MOCK_DATA_ON_INITIALISE=true
 # Used only in development builds
 GENERATE_MOCK_DATA_ON_RESTART=false
 
+# Generate one LLM summary that is used for all questions
+GENERATE_AI_MOCK_DATA=true
```
kaljarv, line 60: Maybe we could prefix all the llm-related env vars by `LLM_`, thus `LLM_GENERATE_MOCK_DATA_ON_INITIALISE`.

.env.example
```diff
@@ -56,7 +56,12 @@ GENERATE_MOCK_DATA_ON_INITIALISE=true
...
+
+OPENAI_API_KEY=""
```
kaljarv, line 62: Maybe make this generic so that if we implement further providers, we can use the same var: `LLM_API_KEY`.

.env.example
```diff
@@ -56,7 +56,12 @@ GENERATE_MOCK_DATA_ON_INITIALISE=true
 # Used only in development builds
 GENERATE_MOCK_DATA_ON_RESTART=false
 
+# Generate one LLM summary that is used for all questions
```
kaljarv, line 59: Generate LLM mock data during mock data generation. Only applicable in development builds and if either of the GENERATE_MOCK_DATA variables is true

backend/vaa-strapi/src/constants.ts
```diff
@@ -12,6 +12,12 @@ export const generateMockDataOnInitialise = process.env.GENERATE_MOCK_DATA_ON_IN
 export const generateMockDataOnRestart =
   (process.env.GENERATE_MOCK_DATA_ON_RESTART === 'true' && nodeEnv === 'development') || false;
 
+/**
+ * When set to true, generates a single llm-response that will be used for every answer
```
kaljarv, line 16: When set to true, generates LLM mock data.

(Let's move the detailed description to the actual implementation.)

backend/vaa-strapi/src/functions/generateMockData.ts
```diff
@@ -668,6 +675,65 @@ async function createCandidateUsers() {
   });
 }
 
+async function generateMockLLMSummaries() {
```
kaljarv, line 678:
```
/**
 * Generates a single llm-response that will be used for every answer.
 */ 
```

packages/llm/README.md
```diff
@@ -0,0 +1,54 @@
...
+npm install @openvaa/llm
+```
+
+2. Create an instance of the OpenAI provider:
+
+```typescript
+import { OpenAIProvider } from '@openvaa/llm';
+
+const llm = new OpenAIProvider({
+  apiKey: 'your-api-key', // Optional if OPENAI_API_KEY env var is set
```
kaljarv, line 29: I think it might be better to not use the env vars within the module but expect the server to always pass the API key.

In that case, we'd also need to read the key into envs and constants in Strapi.

packages/llm/README.md
```diff
@@ -0,0 +1,54 @@
...
+  - Token counting
+  - Usage statistics and cost estimation
+  - Context window management
+
+## Quick Start
+
+1. Install the package:
+
+```bash
+npm install @openvaa/llm
```
kaljarv, line 20: We use yarn so `yarn workspaces focus @openvaa/llm`

packages/llm/README.md
```diff
@@ -0,0 +1,54 @@
...
+
+```typescript
+import { Message, Role } from '@openvaa/llm';
+
+const response = await llm.generate([
+  new Message(Role.SYSTEM, 'You are a helpful assistant'),
+  new Message(Role.USER, 'Hello!')
+]);
+
+console.log(response.content);
```
kaljarv, line 44: console.info

packages/llm/src/llm-providers/__tests__/openai-provider.test.ts
kaljarv, line 1: Can you keep the test files in the same folder as the targets.

packages/llm/src/llm-providers/openai-provider.ts
```diff
@@ -0,0 +1,123 @@
...
+export class OpenAIProvider extends LLMProvider {
+  private model: string;
+  private openai: OpenAI;
+  public readonly maxContextTokens: number;
+
+  constructor(options: { model?: string; apiKey?: string; maxContextTokens?: number } = {}) {
+    super();
+    this.model = options.model || 'gpt-4o-mini';
+    this.maxContextTokens = options.maxContextTokens || 4096;
+    const apiKey = options.apiKey ?? process.env.OPENAI_API_KEY;
```
kaljarv, line 13: See note on env above.

packages/llm/src/llm-providers/llm-provider.ts
```diff
@@ -0,0 +1,79 @@
...
+  }>;
+
+  /**
+   * Maximum number of tokens that can be processed in a single request
+   */
+  abstract get maxContextTokens(): number;
+
+  /**
+   * Calculates how many messages can fit within the context window
+   * while maintaining good model performance
```
kaljarv, line 73: Please don't break lines in comment paragraphs. We expect the IDE to handle line wrapping when needed.

packages/llm/src/llm-providers/llm-provider.ts
```diff
@@ -0,0 +1,79 @@
+export type Role = 'system' | 'user' | 'assistant' | 'developer';
+
+export class Message {
+  public role: Role;
```
kaljarv, line 4: `public` can be omitted from these.

packages/llm/src/llm-providers/llm-provider.ts
```diff
@@ -0,0 +1,79 @@
...
+  public content: string;
+
+  constructor({ role, content }: { role: Role; content: string }) {
+    this.role = role;
+    this.content = content;
+  }
+}
+
+export class UsageStats {
+  constructor(
```
kaljarv, line 14: Please convert to named params.

packages/llm/src/llm-providers/llm-provider.ts
```diff
@@ -0,0 +1,79 @@
...
+   * Currently returns 0 as cost calculation depends on the specific model and pricing.
+   * Override this in provider-specific implementations if needed.
+   */
+  get estimatedCost(): number {
+    return 0;
+  }
+}
+
+export class LLMResponse {
+  constructor(
```
kaljarv, line 31: Please convert to named params.

packages/llm/src/llm-providers/llm-provider.ts
```diff
@@ -0,0 +1,79 @@
...
+ * Implement this class to add support for new LLM providers.
+ */
+export abstract class LLMProvider {
+  /**
+   * Generates a response from the LLM based on the provided messages
+   * @param messages Array of messages to send to the LLM
+   * @param temperature Controls randomness in the response (0-1)
+   * @param maxTokens Optional maximum number of tokens to generate
+   */
+  abstract generate(messages: Array<Message>, temperature: number, maxTokens?: number): Promise<LLMResponse>;
```
kaljarv, line 57: Please convert to named params.

kaljarv (changes requested):
Looks very nice! Just some minor tweaks.

packages/llm/src/llm-providers/openai-provider.ts
```diff
@@ -0,0 +1,132 @@
+import OpenAI from 'openai';
+import { LLMProvider, LLMResponse, Message, UsageStats } from './llm-provider'; // Assuming the previous code is saved in another file
+
+export class OpenAIProvider extends LLMProvider {
+  private model: string;
+  private openai: OpenAI;
+  public readonly maxContextTokens: number;
+
+  constructor(options: { model?: string; apiKey: string; maxContextTokens?: number }) {
```
kaljarv, line 9: For consistency, maybe destructuring with default assignment could be used instead of `options` here as well and move the apiKey throw before calling super.

packages/llm/src/llm-providers/openai-provider.ts
```diff
@@ -0,0 +1,132 @@
...
+      if (error instanceof Error) {
+        throw new Error(`OpenAI API error: ${error.message}`);
+      }
+      throw new Error('Unknown error occurred while calling OpenAI API');
+    }
+  }
+
+  /**
+   * Estimates the number of tokens in a text string.
+   * Note: This is a simple approximation. For production use,
```
kaljarv, line 81: Could you remove the forced line breaks from comments where they do not separate paragraphs.

packages/llm/src/llm-providers/openai-provider.ts
```diff
@@ -0,0 +1,132 @@
...
+   */
+  async countTokens(text: string) {
+    return {
+      tokens: Math.ceil(text.length / 4)
+    };
+  }
+
+  /**
+   * Calculates how many messages can fit within the context window.
+   * Uses a conservative estimate of average tokens per message to ensure
```
kaljarv, line 92: Ditto.

packages/llm/src/llm-providers/openai-provider.ts
```diff
@@ -0,0 +1,132 @@
...
+   */
+  async fitCommentArgsCount(): Promise<number> {
+    const averageTokensPerMessage = 50;
+    return Math.floor(this.maxContextTokens / averageTokensPerMessage);
+  }
+}
+
+function mapToMessageParam(message: { role: string; content: string }): OpenAI.ChatCompletionMessageParam {
+  const role = message.role.toLowerCase();
+
```
kaljarv, line 103: `content` could be destructure here and just passed as `{ role: 'system', content }` below.

kaljarv (approved):
Hi! Looks very good. I just added a couple of stilistic nitpicky notes. 🤓 

Before merging, I think you could just squash all of the other commits after the first (init) into one since this is the first such for this package.

PR #673, opened 2025-01-28
feat: ai question info UI

### LINKED ISSUE #646: Voter UI popup

aleksilassila:
@kaljarv let me know if there are any changes you'd like to see, I will make some changes before changing this from draft to pr

closes #646


frontend/src/lib/components/questions/QuestionInfo.svelte
kaljarv, line 1: Can we actually change the component hierarchy a bit:

1. Keep the old `QuestionInfo` as it is but rename it to `QuestionBasicInfo`.
2. Turn this component a`QuestionExtendedInfo` component which only has the content, i.e., everything inside `Drawer` but not the `Drawer` itself.
  - Removing `title`, `onCollapse`, `onExpand` props
3. Create a new `dynamic-component` called `QuestionExtendedInfoLauncher` (or smth) that will render the button for opening the drawer, access `AppContext` to use the modal queue and handle showing it.
  - For convenience you prob need a non-exported utility component in the same folder that wraps `QuestionExtendedInfo` in `Drawer` so you can pass it to the queue.
  - The new component may take as its input the full `Question` object and destructure the necessary properties from it.
  - For the title of the button add a new translation e.g. `components.questionExtendedInfoLauncher.title` with values "Learn more" / "Tutustu aiheeseen" / "Upptäck mer". 

Lmk if this doesn't make sense or if you can think of a better solution.

frontend/src/lib/contexts/app/modal/modalStore.type.ts
```diff
@@ -0,0 +1,21 @@
...
+ * The popups are displayed by the layout initiating `AppContext` one at a time.
+ */
+export type ModalStore = Readable<
+  { component: ModalComponent<Record<string, unknown>>; props: Record<string, unknown> } | undefined
+> & {
+  /**
+   * Push a new popup component to the stack.
+   * @param component - A component exposing the `onClose` prop.
+   */
+  push: <T extends Record<string, unknown>>(component: ModalComponent<T>, props?: T) => void;
```
kaljarv, line 15: Change T to TProps, bc we adhere to [TS style guide](https://mkosir.github.io/typescript-style-guide/).

frontend/src/routes/[[lang=locale]]/(voters)/(located)/questions/[questionId]/+page.svelte
```diff
@@ -252,7 +255,27 @@ Display a question for answering.
 
     <!-- !videoProps && -->
     {#if info && info !== ''}
```
Comment (kaljarv) [inline frontend/src/routes/[[lang=locale]]/(voters)/(located)/questions/[questionId]/+page.svelte:257]: Could you conditionally show `QuestionBasicInfo` or the the new thing based on a new setting. See the `Settings` section I added yesterday in the [epic](https://github.com/orgs/OpenVAA/projects/9/views/1?pane=issue&itemId=85045077&issue=OpenVAA%7Cvoting-advice-application%7C627).

Also, the `{#if info && info !== ''}` condition should apply to the basic version only, for the extended one check whether `info` is truish OR `infoSections` is non-empty and has length > 0.

frontend/src/routes/[[lang=locale]]/(voters)/(located)/questions/[questionId]/+page.svelte
```diff
@@ -264,7 +287,8 @@ Display a question for answering.
           {disabled}
           {question}
           {selectedId}
-          variant={customData?.vertical ? 'vertical' : undefined}
+          variant={// @ts-ignore
```
Comment (kaljarv) [inline frontend/src/routes/[[lang=locale]]/(voters)/(located)/questions/[questionId]/+page.svelte:290]: We shouldn't need ts-ignore here. Do you get an error?

kaljarv (changes requested):
Looks good 👍 , but can we change the component hierarchy a bit.

frontend/src/lib/components/modal/drawer/Drawer.svelte
```diff
@@ -75,4 +76,11 @@ See the [`<Modal>` component](../Modal.svelte) documentation for more informatio
   container={DrawerContainer}>
   <slot name="actions" slot="actions" />
   <slot />
+  <Button
+    type="button"
+    variant="icon"
+    text="close"
+    class="!absolute bottom-16 right-16 rounded-full bg-base-300 p-14"
```
aleksilassila, line 83: Do we have an existing component for buttons with the desired styling? If not, how do you prefer I style this so that it works with dark/light modes? Should I create a new button variant for a floating icon button?

frontend/src/lib/components/modal/Modal.svelte
kaljarv, line 1: Hi!

I wonder if we could achieve the same behaviour without needing to pass components as properties, which makes things a bit too complicated, by doing something like this.

### ModalContainer

Make a new `ModalContainer` component, which contains everything from the old `Modal` except the contents below. It handles all the modal functionality.

```tsx
  <div class="modal-box {boxClass ?? ''}">
    <h2 id={titleId} class="mb-lg text-center">{title}</h2>
    <slot />
    {#if $$slots.actions}
      <div class="modal-action justify-center">
        <slot name="actions" />
      </div>
    {/if}
    <form method="dialog">
      <button class="btn btn-circle btn-ghost btn-sm absolute right-2 top-2">
        <span aria-hidden="true">✕</span>
        <span class="sr-only">{$t('common.closeDialog')}</span>
      </button>
    </form>
  </div>
```

Moreover:
- it would only have one `slot`
- change `dialog.aria-labelledby={titleId}` to `aria-label={title}`
- no `boxClass` prop

### Modal

The old `Modal` would have the same props and slots as before and would in effect render:

```tsx
<ModalContainer {...$$restProps} {title}>
  <div class="modal-box {boxClass ?? ''}">
    <h2 id={titleId} class="mb-lg text-center">{title}</h2>
    <slot />
    {#if $$slots.actions}
      <div class="modal-action justify-center">
        <slot name="actions" />
      </div>
    {/if}
    <form method="dialog">
      <button class="btn btn-circle btn-ghost btn-sm absolute right-2 top-2">
        <span aria-hidden="true">✕</span>
        <span class="sr-only">{$t('common.closeDialog')}</span>
      </button>
    </form>
  </div>
</ModalContainer>
```

### Drawer

Would be built the same way as the new `Modal` above.

### Added bonus

If this works and you have the time and energy, I think we could get rid of the clunky function forwarding currently used in `TimedModal` and which we would need to do with `Modal` and `Drawer` by converting `open/closeModal` functions in `Modal` (hence `ModalContainer`) and adding them to the props types:

```ts
// ModalContainer
export let openModal: $$Props['openModal'] = () { .... }
// Was: export function openModal() { ... }
```

And then in wrappers of `ModalContainer`:

```tsx
export let openModal: $$Props['openModal'];
<Modal bind:openModal>
// Was:
// export function openModal() {
//  _openModal();
// }
// let _openModal: () => void;
// <Modal bind:openModal={_openModal}>
```

frontend/src/lib/components/modal/drawer/Drawer.svelte
kaljarv: There's no such yet, but could you indeed add such a `floating-icon` variant to `Button`. Maybe you can define in that one the placement using margins instead, i.e., `m-16`, and leave the absolute classing to the user of the component so that they may choose its placement and get it right by, e.g., applying `absolute bottom-0 end-0`.

Also, there's a small addition top `Button` not yet merged to main, so can you cherry-pick it and build on top of that: https://github.com/OpenVAA/voting-advice-application/pull/660/commits/4076ce9a6bdf3c66b88b26507ba39dd8d2b3e735

frontend/src/lib/components/modal/Modal.svelte
aleksilassila: That seems like a good idea. Is this what you envisioned? https://github.com/OpenVAA/voting-advice-application/pull/673/commits/7ce1b53188a836ee22c2bacf49ed9e62b614edd3

kaljarv: Thanks! It looks good now.

The only thing is, I guess you could remove the `actions` slot from `Drawer` because it might not work properly now that the DaisyUI `modal-box` class is not used in its containing div.

Is the PR otherwise ready for a full review or are you still gonna add something?

aleksilassila: 👍 it's ready for review now

kaljarv (changes requested):
Looks very good! Just a number of small tweaks.

frontend/src/lib/components/button/Button.type.ts
```diff
@@ -21,7 +21,7 @@ export type ButtonProps = LinkOrButtonElementProps & {
   /**
    * Type of the button, which defines it's appearance. @default 'normal'
    */
-  variant?: 'main' | 'icon' | 'normal' | 'responsive-icon' | 'secondary' | null;
+  variant?: 'main' | 'prominent' | 'icon' | 'normal' | 'responsive-icon' | 'secondary' | 'floating-icon' | null;
   /**
    * Position of the icon in the button. Only relevant if `icon` is not `null` and `variant` is not `icon`. Note that `top` and `bottom` are not supported if `variant='main'`. @default 'right' if `variant='main'`, otherwise `left`
```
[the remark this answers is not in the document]
aleksilassila: Sorry, what do you mean here

frontend/src/lib/components/questions/QuestionExtendedInfo.type.ts
```diff
@@ -0,0 +1,15 @@
...
+  /**
+   * The info content to show as a plain or HTML string.
+   */
+  info: string;
+  /**
+   * Additional expandable info sections shown as plain or HTML strings.
+   */
+  infoSections?: Array<{
+    title: string;
+    content: string;
```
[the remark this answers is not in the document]
aleksilassila: My thinking was that the ui component would be agnostic to the backend logic of not showing certain sections, and the filtering would be done on higher level, but we can do this, too. Could you elaborate, where do you want to filter `infoSections` exactly?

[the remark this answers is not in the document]
kaljarv: Ah, alright, that sounds like a good solution. Then it should be done in `UniversalDataProvider.getQuestionData`. Everything passes through that in the end.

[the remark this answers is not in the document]
kaljarv: @aleksilassila, could you still do this change? Maybe at the same time align the naming in the CustomData type even though this component could be agnostic to that. 

```ts
type QuestionInfoSection = {
  title?: string;
  content?: string; // Was text
  visible?: boolean;
};
```

kaljarv (changes requested):
Sorry, there was still one thing about `visiblity`.

frontend/src/lib/components/questions/QuestionExtendedInfo.svelte
```diff
@@ -0,0 +1,47 @@
...
+<script lang="ts">
+  import { sanitizeHtml } from '$lib/utils/sanitize';
+  import { Expander } from '../expander';
+  import type { QuestionInfoProps } from './QuestionExtendedInfo.type';
+
+  type $$Props = QuestionInfoProps;
+
+  export let info: $$Props['info'];
+  export let infoSections: $$Props['infoSections'] = [];
+  $: visibleInfoSections = infoSections?.filter(
```
kaljarv, line 31: If the check is left here, however, (see the comment in the main discussion) I think you should be able to do this without the `section is Required<QuestionInfoSection>` part. If the linter complains, then the typing for the property must be off…

kaljarv (approved):
Thanks for the changes! Could you check the comment here and the related one in the main discussion and decide what to do. Then it's good to go, unless you want me to take another look.

Issue #627, opened 2024-10-29
Interactive chat for explaining terms and question


kaljarv:
# Primary user story

As a voter unfamiliar with a VAA statement,
I want to receive answers to questions I have about the statement,
so that I can make a more informed choice when answering.

# Considerations

## Structured or free chat?

It is probably easier for a user to start the chat if there are preprovided questions or highlighted terms they can tap on instead of them having to formulate a freeform query. Such a structured approach provides the additional benefit that the questions and answers can precomputed, approved and edited by the administrator.

If the chat is structured, it may be useful to precompute follow-up questions up to some level as well.

However, voters might have a questions about the statements that are not included in the precomputed set, and in these cases a free chat would be useful.

# Subtasks

1. Initial planning
  - Which technologies to use?
  - Which of the tech is shared between the other ideas?
  - Which data sources are needed?
  - Is the AI model interacted with in real-time or do we use canned results?
  - Make a sketch of the ideal interaction? E.g. a script of the chat or a rough draft of the visualisation
2. Prototyping
  - Make a UI sketch in [Figma](https://www.figma.com/design/3LDNZWOxuhhVaVPAOApfOK/Voting-Advice-Application?node-id=4169-15907&t=o7Z4IrwYPkeGbBeH-1)
  - Test the suggested workflows using real data and real models
3. Development
  - Structured 'canned' chat
  - Generate:
    - Introduction to question
    - Explanations for all terms in the statement and the introduction
    - What is the current situation
    - Ability to add more generatable items
  - Localizable
  - Voter UI
    - Common with #628
4. Optional admin UI development
  - Edit introductions and 'current situation' summaries
  - Edit, delete and add term explanations
    - Automatically link these to the terms in the introductions without explicit a-elements?
  - Localizable
  - Perhaps just an export/import of tabular or JSON data

# Update 2025-01-28

## Strapi endpoints

- Define what new Strapi endpoints are needed (read/write)
- Write functions will have to run on server only to be able to use an API key stored in `env` variables (or in the future use a new authenticated admin user type)

## Settings

An example of the settings to add for the new functions.

For now it's okay if these are only added to `packages/app-shared` and not implemented in Strapi’s `App Settings`.

```ts
// In app-shared: dynamicSettings.ts
interface DynamicSettings {
  // ... existing stuff
  questions: {  
    // ... existing stuff
    interactiveInfo?: {
      /**
       * Default `false`.
       */
      enabled?: boolean;
      // Possible other settings
[...]

PR #689, opened 2025-02-11
feat: Add example prompts and LLM output template

### LINKED ISSUE #659: #627 Prompts for most common questions
Questions from the second page (see Figma).

Create prompt format for most common questions.

edthu:
## WHY:

Fixes #659 

Added files with default prompts and a default data-format for LLM-output. [Usage](https://github.com/OpenVAA/voting-advice-application/compare/feat/llm...feat/llm-prompts?expand=1#diff-c0b54ebb2e027300f61beb69562063723c1d1fadb154d0f5ba676337f0ba5035R683)

## Questions about this feature and the future

The idea is for these to be used in editing prompts and questions as described in #627 4.

The implementation of that will be done in the next sprint (if it is chosen from the items) but just to start planning for it how should that feature be implemented? There was some kind of logging in panel for the frontend. Should the admin-ui be there? Should the other features have some options in there?

@kaljarv


backend/vaa-strapi/src/functions/generateMockData.ts
```diff
@@ -678,34 +680,22 @@ async function createCandidateUsers() {
 async function generateMockLLMSummaries() {
   const OPENAI_API_KEY = process.env.OPENAI_API_KEY;
   try {
+    const promptPath = join(__dirname, 'prompts', 'defaultPrompt.json');
```
kaljarv, line 683: We could import these the same way as other `json` data files in this file and make sure there's always something.

backend/vaa-strapi/src/functions/generateMockData.ts
```diff
@@ -678,34 +680,22 @@ async function createCandidateUsers() {
 async function generateMockLLMSummaries() {
   const OPENAI_API_KEY = process.env.OPENAI_API_KEY;
```
kaljarv, line 681: Could you add this to `src/constants.ts` and import it from there? Note also that the name was just changed to `LLM_ OPENAI_API_KEY`.

backend/vaa-strapi/src/functions/generateMockData.ts
```diff
@@ -678,34 +680,22 @@ async function createCandidateUsers() {
 async function generateMockLLMSummaries() {
```
kaljarv, line 680: Now that we're actually using the API to generate, refactor admittedly messy `generateMockData` logic so that we only generate the summaries once – and if `generateMockDataOnRestart` is true use the same summaries (the questions stay the same anyway).

kaljarv (commented):
Some random notes.

kaljarv:
As for "There was some kind of logging in panel for the frontend. Should the admin-ui be there? Should the other features have some options in there?"

The features options should be in the frontend admin UI, but it's optional to enable editing these prompts there.

Actually, now that I think about it, maybe these templates could just be part of `app-shared: DynamicSettings` with their own key in which case they could be either edited directly in the files or via Strapi…

backend/vaa-strapi/src/functions/generateMockData.ts
edthu: Settings moved to DynamicSettings in [7d42307](https://github.com/OpenVAA/voting-advice-application/pull/689/commits/7d42307b7d213aee2ba726dbcd03b2523a74a3ee)

backend/vaa-strapi/src/functions/generateMockData.ts
edthu: Done

backend/vaa-strapi/src/functions/generateMockData.ts
edthu: There is only one summary generated during initialization and restart that is used for all questions so the amount of requests is pretty small unless this is restarted a lot. Does this need to be done? If this has already cost more than I estimated then saving the prompt sounds like a good idea :D

kaljarv: Oh, true, we can leave as it is then!

packages/app-shared/src/settings/dynamicSettings.type.ts
```diff
@@ -203,6 +203,13 @@ export type DynamicSettings = {
     disallowSelection?: boolean;
   };
   underMaintenance?: boolean;
+  /**
+   * Default prompt and answer format for generating llm-summaries from question descriptions. The llm-generated answers needs to be in a JSON-format and needs to be specified in the answerFormat
+   */
+  LLMSettings: {
```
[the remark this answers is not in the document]
edthu: changed in [a7318ff](https://github.com/OpenVAA/voting-advice-application/pull/689/commits/a7318ff812add6e634548f419325aff93191bf01)

PR #703, opened 2025-02-19
feat: de novo candidate registration UI (enhancement)

shmeleva:
## WHY:

Fixes https://github.com/orgs/OpenVAA/projects/10?pane=issue&itemId=98163487

* Verify `audience` and `issuer` ID token claims.
* Support decryption JWK rotation.
  * Before: `IDENTITY_PROVIDER_ENCRYPTION_PRIVATE_KEY` held a single JWK.
  * After: `IDENTITY_PROVIDER_ENCRYPTION_PRIVATE_KEY` has been replaced with `IDENTITY_PROVIDER_DECRYPTION_JWKS`, which stores a list of JWKs. When decrypting an ID token, the key ID (`kid`) from the protected header is used to look up the correct key from the list.
* Generate a code verifier and a code challenge.
  * Store the code verifier in `localStorage`.
  * Send the code challenge and challenge method to the authorization endpoint.
  * Send the code verifier to the token endpoint.
  * More: https://developer.signicat.com/docs/eid-hub/oidc/advanced-security/proof-key-for-code-exchange-pkce/, https://datatracker.ietf.org/doc/html/rfc7636

**Should I add some tests?**


kaljarv:
@shmeleva, I rebased this now on `main`. Could you check that everything's still intact?

As for tests, I guess it would be nice if you could add unit tests for the utilities you added. For e2e tests, we could add one that can be optionally run if a current subscription to Signicat exists. I guess trying to mock everything would be too complicarted, right?

PR #717, opened 2025-03-07
feat: Make performance-optimized alternative to /api/nominations and /api/question-categories

shmeleva:
## WHY:

Fixes https://github.com/orgs/OpenVAA/projects/10?pane=issue&itemId=89239431&issue=OpenVAA%7Cvoting-advice-application%7C504

Strapi already adds `BTREE` indexes on all primary and foreign key columns and document ID columns in the relevant tables:
`elections`, `constituencies`, `files_related_mph`, `files`, `parties`, `nominations`, `nominations_election_lnk`, `nominations_constituency_lnk`, `nominations_candidate_lnk`, `candidates`, `candidates_party_lnk`, `nominations_party_lnk`. There isn't much room for further optimization in this area.

However, Strapi’s SQL query generation is somewhat suboptimal. It makes duplicate joins on the same tables, performs multiple `SELECT` queries to support pagination (which we do not benefit from at the moment), and there might be more.

There is a Redis caching plugin available for Strapi 4, but not for Strapi 5.

A custom `/api/nominations/with-relations` API endpoint has been added to provide a performance-optimized alternative to the default Strapi `/api/nominations` endpoint.
* In [5cd25fb](https://github.com/OpenVAA/voting-advice-application/pull/717/commits/5cd25fb694d6ef3f718bb36ca6e3026308cd1643), all the same columns as in the original request were selected. However, in [508a32b](https://github.com/OpenVAA/voting-advice-application/pull/717/commits/508a32b17724e22f430571fbf880c2106dce63af) some columns were removed, keeping only those required by the `StrapiNominationData` type to make the response more compact.
* Filtering arguments follow the same format as before, but nominations can only be filtered by election and constituency.

A custom `/api/question-categories/with-relations` API endpoint has been added to provide a performance-optimized alternative to the default Strapi `/api/question-categories` endpoint.
* Filtering arguments follow the same format as before, but question categories can only be filtered by elections.
* This still needs to be tested a bit more to ensure that filtering works as before ❗ 

### DIFF
--- a/backend/vaa-strapi/src/api/nomination/controllers/nomination.ts
+++ b/backend/vaa-strapi/src/api/nomination/controllers/nomination.ts
@@ -25,6 +25,9 @@ export default factories.createCoreController('api::nomination.nomination', ({ s
    *       $in?: string[];
    *     };
    *   };
+   *   unconfirmed?: {
+   *     $ne?: 'true' | 'false';
+   *   };
    * };
    * ```
    *
@@ -64,6 +67,9 @@ export default factories.createCoreController('api::nomination.nomination', ({ s
             $in?: Array<string>;
           };
         };
+        unconfirmed?: {
+          $ne?: 'true' | 'false';
+        };
       };
     };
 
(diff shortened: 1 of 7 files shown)


backend/vaa-strapi/src/api/nomination/controllers/nomination.ts
```diff
@@ -4,4 +4,217 @@
...
+              'color', parties.color,
+              'name', parties.name,
+              'shortName', parties.short_name,
+              'info', parties.info,
+              'colorDark', parties.color_dark,
+              'answers', parties.answers,
+              'image', _images.data
+            ) data
+          from parties
+          left join _images on _images.related_id = parties.id
```
kaljarv, line 155: This part seems to add random images (maybe because it's referencing `id` instead of `documentId`) to parties. I just removed from the current version, bc there are no party logos anyway 😄

kaljarv:
As of the disk cache implementation, these changes are no longer needed.

PR #734, opened 2025-03-18
feat: implement question ordering with factor analysis

### LINKED ISSUE #656: [631] Integrate factor-based ordering algorithm into frontend

MemeSlayer27:
- Add new @openvaa/factor-analysis package with polychoric correlation and factor analysis
- Implement question ordering based on information gain from factor loadings
- Add UI for multiple question selection with "choose for me" option
- Add settings for dynamic ordering and question selection count
- Track shown questions in questionBlockStore
- Add comprehensive tests and documentation

Fixes #656, #695, #694

image


kaljarv (approved):
Hi!

I added only two random details in the comments, but here are some general answers to the questions you asked elsewhere.

### `QuestionOrderer`

On `/frontend/src/routes/[[lang=locale]]/(voters)/(located)/questions/[questionId]`, could you check if we could move the `getNextQuestion` and (at least partially) the `getNextQuestionChoices` directly into the `selectedQuestionBlocks` store (perhaps even as derived stores instead of functions), which may require some refactoring to accomodate these new features.

Moreover, the `QuestionOrderer` instance could live within it or as a separate object in `VoterContext`, derived from `factorLoadings`. If I understood correctly, the `QuestionOrderer` only needs to be reinstantiated if `factorLoadings` changes, right?

### `FactorLoadings`

The `data` package is meant for the core election data, from which I would still exclude the factor loadings, especially since they don't need that much access to the other data objects. These would be more akin to `AppCustomization` which is defined within the `AppContext`.

Could you move the code from `data` to `$lib/contexts/voter/factorLoadings` (with appropriated refactoring).

### Factors with multiple `Election`s

It's quite hard to strictly separate questions by election, because some of the questions may be shared between elections whilst others apply to only a specific one.

Would it be possible to use all questions regardless of which election they apply to when calculating the factors? And then when using them for question ordering somehow compute the residual entropy based on the subset of questions in the currently selected elections?

If keeping with the election-speficif factors, maybe this kind of procedure would work when defining the question order:

1. compute the information gain for each question for each election (with it being zero for questions that don't apply to the specific election)
2. for ordering, use either
   - a (normalized if isn't already) sum of the gains
   - the gain for the election that currently has the most residual entropy
3. compute the reliability score from the election that currently has the most residual entropy

### Other details

Empty `Answer`s can be found by:

1. `entity.answers[questionId]?.value == null`
2. But the more reliable options is to use `question.ensureAnswer(entity.answer) == null` because it also checks that `value` is valid for that specific `Question`.

The types of the other matchable questions other than Likert ones can be found here:

- `/packages/data/src/objects/questions/variants/singleChoiceCategoricalQuestion.ts`
- `/packages/data/src/objects/questions/variants/numberQuestion.ts`
- `/packages/data/src/objects/questions/variants/booleanQuestion.ts`

(The `PreferenceOrder` type is not yet implemented.)

It would probably make sense to build a method in the `factor-analysis` package itself for accepting inputs using the objects (`Question` & `Answer`) defined in the `data` package.

PR #735, opened 2025-03-20
feat: Disk cache

shmeleva:
## WHY:

Fixes [Issue link]


kaljarv (commented):
We could try to move the cache one step higher up in the call chain so that it'd be `UniversalAdapter` that accesses the cache.

This would entail adding a `doFetch(url, request, options: { disableCache?: boolean; } )` or smth method to `UniversalAdapter` and move the following part from `apiFetch` (from both `ApiRouteAdapter` and `StrapiAdapter`) there along these lines:

```ts
const useCache = staticSettings.cacheEnabled && !disableCache && request.method === 'GET';
const response = await this.fetch(useCache ? cachifyUrl(url) : url, request);
if (!response.ok) {
  const { error } = await response.json();
  throw new Error(`Error with doFetch: ${response.status} (${response.statusText ?? '-'}) • ${url}`, {
    cause: error?.message
  });
}
return response;
```

`ApiRouteAdapter.apiFetch` and `StrapiAdapter.apiFetch` would then internally call `this.doFetch` passing any `disableCache` args from their own getter implementations.

One additional thing to consider would also be moving `if (authToken) request = addHeader(request, 'Authorization', `Bearer ${authToken}`);` to the new-fangled `doFetch` as well. That way we could also automatically disable the cache if auth is set, although we'd probably want to check for the raw header as well…

And one more would be to disable the cache when in the development environment – or even better using a `PUBLIC_USE_CACHE_IN_DEV` env var (false by default) to set that.

kaljarv:
Replaced by #789

PR #777, opened 2025-05-24
feat: further improvements from nuorten vaalikone 2025

kaljarv:
A further collection of improvements that were added when deploying Nuorten Vaalikone 2025 and its educational mode (the actual implementation of the game mode, however, is contained in a different PR).


frontend/src/lib/contexts/voter/matchStore.ts
```diff
@@ -53,13 +63,39 @@ export function matchStore({
               ? removeDuplicates(questions.map((q) => q.category))
               : undefined;
 
-            const matches = algorithm.match({
+            // Possibly impute parent entity answers
+            let proxies: Array<MatchingProxy<AnyNominationVariant>> | undefined;
```
kaljarv, line 67: We use these rather unwieldy proxies to store the imputed answers for parties because we do not want to overwrite the actual parties’ answers. Even if there wouldn't be any, we would run into trouble when switching between elections or constituencies, because the proxied answers are based on the a party’s candidates in a given election and constituency.

Optimally, such imputations would already be precalculated in the database with equal granularity but that remains to be done later.

frontend/src/lib/contexts/voter/nominationAndQuestionStore.ts
```diff
@@ -67,21 +69,29 @@ export function nominationAndQuestionStore({
                 entityType,
                 type: QUESTION_CATEGORY_TYPE.Opinion
               });
+              // Check that the entities have answers to all opinion questions
```
kaljarv, line 72: Ideally, this check would be done in the backend, but it's better left after the backend migration away from Strapi.

frontend/Dockerfile
```diff
@@ -36,7 +36,7 @@ COPY --from=shared /opt/frontend frontend
 
 FROM frontend AS development
 
-WORKDIR /opt
+WORKDIR /opt/frontend
```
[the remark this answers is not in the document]
kaljarv: And i'ts changed now in https://github.com/OpenVAA/voting-advice-application/pull/777/commits/45d6b3881da8b2ab64b84b4670a79b682ac2cf73

PR #799, opened 2025-08-24
fix: `instanceof` checks

### LINKED ISSUE #787: fix: `instanceof` checks for custom classes
These may break when code is bundled and should be circumvented.

kaljarv:
## WHY:

Fixes #787 

Convert all istanceof checks used for custom classed defined in the
packages to use proprietary type guard utilities.

Instanceof does not work reliably across packages due to name changes
in bundling.


packages/data/src/utils/typeGuards.ts
kaljarv, line 1: This format of type guards is now used for the submodules.

frontend/src/lib/components/entityFilters/EntityFilters.svelte
```diff
@@ -16,15 +16,26 @@ Show filters for entities. This component and the individual filter components o
...
 
   export let filterGroup: $$Props['filterGroup'];
   export let targets: $$Props['targets'];
+
+  const { t } = getComponentContext();
+
+  /** Type params cannot be used in the HTML part */
+  function _isEnumeratedFilter(filter: unknown) {
+    // TODO[Svelte 5]: Check if needed
+    return isEnumeratedFilter<MaybeWrappedEntityVariant, AnyEntityVariant>(filter);
```
kaljarv, line 37: Here we need to specify the type params so that `<EnumeratedEntityFilter>` accepts the filters. Note that types and type params cannot be used in the HTML part, hence the wrapper.

PR #863, opened 2026-08-17
ship(v0.2) 01/12 — move frontend/ and docs/ under apps/ (renames only, zero lines)

kaljarv:
**Stack position: 1 of 12.** Base: `main`. Head: `ship/v0.2-akita-01a-layout-move`.
**1,316 files changed, 0 insertions, 0 deletions.** Every one of them is a rename.

---

   are not. See "Why this PR is red" below for exactly which checks fail here and why.
   zero files. The proof is recorded in the phase record alongside the partition.
   branch targeting whichever of the development branch or the stack is easier at the time. Nothing
   about this stack's shape constrains where a fix goes.

**No file is shown to you twice in two different content versions.** The partition was built so that
every file appears in exactly one slice's diff, in its final state. You will never review a version
that a later PR in the stack undoes. (Verified: every off-diagonal cell of the twelve-by-twelve
pairwise overlap matrix is `0`, and the twelve slices plus an empty remainder reconstruct the target
tree exactly.)

**Only this PR targets `main`.** Every other PR in the stack targets the branch below it, so GitHub
shows you each one's own diff and nothing else.

## What you are looking at

The monorepo layout move: `frontend/` → `apps/frontend/` and `docs/` → `apps/docs/`.

**Paths change. Contents do not.** This commit was reconstructed specifically so that it is a pure
rename set — every output blob is the identical object from the base commit, re-pointed at a new path.
Measured on this commit:

| Measurement | Value |
| --- | --- |
| `git show -M --shortstat` | `1316 files changed, 0 insertions(+), 0 deletions(-)` |
| File-status taxonomy at `diff.renameLimit=1` | **`1316 R`** — and `0 A`, `0 M`, `0 D` |

That taxonomy is asserted at `diff.renameLimit=1`, the maximally hostile setting — the one at which
git is *least* willing to call something a rename. It still reports all 1,316 as renames, so the
result does not depend on a generous rename-detection threshold. A `D` or an `A` in that set would
mean this commit had lost or invented a file; there are none of either.

**So the review here is a single question: are these the right paths?** There is no code to read. It
is a large file count and a trivial diff, and those two facts are not in tension — a rename consumes
a *file* slot and no *line* budget.

Content changes to these same files arrive later in the stack, as ordinary edits at their new paths.
That separation is the entire purpose of doing the move as its own PR: without it, roughly 1,300 files
would render as delete/add pairs inside a content PR, and the real changes would be invisible inside
64,860 added and 64,860 deleted lines that mean nothing. (That is exactly what
`git show --no-renames` reports for this commit: `2632 files changed, 64860 insertions(+),
64860 deletions(-)`. It is the add-plus-delete *rendering* of a move, not a review cost.)

## Why this PR is red, and why that is expected

`.github/workflows/main.yaml` triggers on pull requests whose base is `main`. This is the only PR in
the stack whose base is `main`, so **this is the only PR in the stack that fires CI at all**. On the
other eleven, "no checks ran" is correct behaviour and not a misconfiguration.

All three jobs in that workflow fail here, and **all three fail at the same step with the same error**
— as observed on the run this PR actually triggered:

| Job | Fails at | Error |
| --- | --- | --- |
| `frontend-and-shared-module-validation` | step 3, `Setup Yarn 4.6` | `YN0028` |
| `backend-validation` | step 3, `Setup Yarn 4.6` | `YN0028` |
| `e2e-tests` | step 4, `Setup Yarn 4.6` | `YN0028` |

Note *which* step that is. The dependency install happens **inside** `Setup Yarn 4.6` —
`threeal/setup-yarn-action@v2` installs dependencies itself as part of its caching — so the workflow's
own, more obviously named `Install all dependencies` step is never reached. It and every lint, format,
type-check, test and build step after it are reported as `skipped`, not as failures.

The error, verbatim from the run:

```
YN0085: │ - @adobe/css-tools@npm:4.4.1, @alloc/quick-lru@npm:5.2.0, @capacitor/android@npm:5.7.8,
          @capacitor/cli@npm:5.7.8, @capacitor/core@npm:5.7.8, and 407 more.
```

Those 412 packages dropping out of the resolution *are* the symptom: they are the dependency closure
of the two workspaces this PR moved out from under the paths the root `package.json` still names.

**What passes:** the repository's CodeQL default-setup checks — `Analyze (javascript-typescript)`,
`Analyze (actions)` and `CodeQL` — all pass on this PR. The red is confined to the three `main.yaml`
jobs above.

**The cause, precisely.** The repository root `package.json` at this commit declares:

```json
"workspaces": ["packages/*", "backend/vaa-strapi", "backend/vaa-strapi/src/plugins/*", "frontend", "docs"]
```

That is unchanged by this PR — and it *has* to be, because this PR is renames only and root
`package.json` is not one of the renamed files. But `frontend/` and `docs/` are now
`apps/frontend/` and `apps/docs/`, and `yarn.lock` at this commit still resolves
`@openvaa/frontend@workspace:frontend` and `@openvaa/docs@workspace:docs`. Two of the five declared
workspaces therefore no longer exist on disk while the lockfile still names them.

What the run shows happening: yarn resolves the workspaces that *do* exist, which is a strictly
smaller set than the lockfile records — hence the 412 dropped packages — and `--frozen-lockfile`
(Yarn 4's `--immutable`) then refuses the resulting lockfile change outright. The install fails
before any lint, format, type-check, test or build step in those jobs is reached, which is why none
of them reports anything.

**This is not confined to this PR.** A third declared workspace, `backend/vaa-strapi`, is deleted by
the next PR in the stack. The root `package.json` and `yarn.lock` are only corrected in
`ship/v0.2-akita-10-root-config` (PR 11 of 12), where `workspaces` becomes
`["packages/*", "apps/*"]`. **Every stack state from this commit up to and including
`ship/v0.2-akita-09-docs` — ten branches — therefore cannot `yarn install`.** That is the stack
contract's point (1) working as designed, not a defect discovered late: the layout move must be
content-free to be readable, and the configuration that describes the layout is content.

If you want a green signal for this body of work, the one that exists is the **full test suite run
against the development branch's tip**, recorded in the phase record. That is the collective proof
the stack is measured against. Per-PR CI on a deliberately partial state is not, and cannot be.

- **1,316 files exceeds GitHub's 300-file single-diff render limit.** Expect the diff view to
  truncate and to need "Load more". Because every entry is a rename with zero line changes, what you
  are loading is a path list, not code.
- **The line budget is zero, so no line-based truncation applies.** A render flag on this PR is a
  file-count flag only.
- GitHub does report renames as renames (`renamed` entries in the diff and in the API), so the file
  list should read as `frontend/x → apps/frontend/x` pairs throughout. If you see delete/add pairs
  instead, something is wrong and it is worth saying so — that would contradict the measurement above.

## Disposition record

Every item of `.agents/code-review-checklist.md` was swept against this slice and given a recorded
verdict — met, fixed, or deliberately deferred with a reason. For this slice the honest result is that
**no content-shaped checklist item can be "met" by a pure rename**, and none is claimed as met: its
twelve cells are nine `NOT-SWEPT` (each with a stated reason — there is no content to sweep) and three
`DEFERRED`.

The three deferrals are real findings about the move, recorded rather than hidden:

- **Path references the move invalidates but does not update** — 269 occurrences across roughly 140
  files still name the pre-move locations. The clearest instance: the repository's own front-page
  `README.md` renders ``, and after this
  move that path no longer exists. Most of the remainder is under `apps/docs/**`, owned by PR 10 of the
  stack, plus three `.github/` files owned by PR 11.
- **Dead code relocated rather than removed** — `apps/frontend/jest.config.json` (this repository
  tests with vitest; `yarn.lock` carries zero `jest@npm` entries) and 89 tracked files under
  `apps/frontend/{android,ios}/` (zero `capacitor` entries in `yarn.lock`) are moved by this commit
  and appear in no later slice's diff.
- **Developers' Guide repository links** invalidated by the move.

Full evidence, with file-and-line citations:
**"Slice 01a — `ship/v0.2-akita-01a-layout-move`"**.

[151-DISPOSITION.md § Slice 01a](https://github.com/OpenVAA/voting-advice-application/blob/ship/v0.2-akita-11-planning/.planning/phases/151-ship-v0-2-akita-review-stack/151-DISPOSITION.md#slice-01a--shipv02-akita-01a-layout-move)

## The stack

`main`.

| **1** | **`ship/v0.2-akita-01a-layout-move`** ← *you are here* | move the frontend and docs trees into `apps/` (renames only) | **1316** | **0** | **0** |
| 4 | `ship/v0.2-akita-03-supabase` | replace the Strapi backend with the Supabase schema, RLS, functions and generated types | 118 | 16257 | 0 |
| 5 | `ship/v0.2-akita-04-dev-seed` | add the dev-seed package that generates deterministic local and E2E data | 162 | 19560 | 0 |
| 6 | `ship/v0.2-akita-05-e2e-tests` | add the Playwright end-to-end suite and its runner configuration | 195 | 23297 | 778 |
| 12 | `ship/v0.2-akita-11-planning` | add the v0.2 planning record and agent configuration | 2287 | 866928 | 100 |

is what a reviewer actually sees.


kaljarv (commented):
Checked.

PR #866, opened 2026-08-17
feat[db]: replace the Strapi backend with the Supabase schema, RLS, functions and generated types

kaljarv:
**Stack position: 4 of 12.** Base: `ship/v0.2-akita-02-shared-packages`. Head: `ship/v0.2-akita-03-supabase`.
**119 files changed, 16,422 insertions, 0 deletions** — all 119 are additions.

---

   zero files.

every off-diagonal cell of the pairwise overlap matrix is `0`, and the slices plus an empty remainder
reconstruct the target tree exactly. Re-verified after the slice above this one was cut: the five cut
slices plus the untouched remainder produce tree `c5b0fecde`, which is the development branch's tree
hash.)

## What you are looking at

The Supabase backend that replaces the Strapi one PR 2 deleted: the schema, its row-level security,
its functions, its Edge Functions, its database tests and its generated TypeScript types.

| Area | Files |
| --- | ---: |
| `apps/supabase/benchmarks/**` — k6 and pgbench harnesses plus their recorded results | 62 |
| `apps/supabase/supabase/**` — schema, migrations, Edge Functions, pgTAP tests, `config.toml` | 46 |
| `packages/supabase-types/**` — the generated `Database` type and its maps | 6 |
| `apps/supabase/{README.md, package.json, scripts/lint-schema.mjs, vitest.config.ts, .gitignore}` | 5 |

By extension: **56 `.sql`**, 39 `.json` (almost all benchmark results), 9 `.ts`, 4 `.sh`.

- **The schema is here twice, and `apps/supabase/README.md` — added by this PR — says which copy the
  database actually reads.** `supabase/schema/` is 21 files and 3,409 readable lines; `supabase/migrations/`
  holds the same schema again, and the migrations are what runs. Concatenating `schema/*.sql` in filename
  order and diffing against `migrations/00001_initial_schema.sql` yields exactly the three hunks
  migrations 00002 and 00003 apply, and nothing else — so the two do agree today. Nothing verifies that
  they keep agreeing, and `config.toml`'s `schema_paths` is empty, so the CLI never reads `schema/` at
  all. That is a real maintenance hazard and it is now written down rather than discoverable only from a
  comment inside a migration.
- **Row-level security is the substance of this PR.** 97 policies across 20 tables. Every policy names an
  explicit role target and every one of the 56 `auth.uid()` / `auth.jwt()` call sites uses the
  `(SELECT auth.…())` scalar-subquery form that lets Postgres hoist it out of the row loop. Counts, not
  samples — see below.
- **Three Edge Functions**, 5 files, 1,097 lines: `invite-candidate`, `send-email`, `identity-callback`
  (the unauthenticated leg of bank login).
- **`apps/supabase/benchmarks/` is 62 of the 119 files** and is a measurement harness — k6 scripts,
  pgbench definitions, and committed result JSON. It has no runtime path into the application and holds
  no credential. If you are budgeting review time, spend it on the 46 files under `supabase/`.

## Why no checks run here

`ship/v0.2-akita-02-shared-packages`, a sibling branch, so **no checks run — and that is correct
behaviour, not a misconfiguration.** Only PR 1 of the stack fires CI at all.

**There is a second reason worth stating, because it is specific to this PR.** The workflow file at
this branch's head is still `main`'s — byte-identical, blob `c2fdcedb2` — and it defines exactly three
jobs: `frontend-and-shared-module-validation`, `backend-validation` and `e2e-tests`. **The
`supabase-tests` job that runs the pgTAP suite does not exist at this head.** It arrives with PR 11,
which is where this branch's own updated workflow file lands. The job called `backend-validation` that
*does* exist here builds and validates `@openvaa/strapi` — the workspace PR 2 deleted. So even setting
the base aside, there is no CI job at this point in the stack that could exercise this PR's contents.

Had the three jobs run, they would have failed at the same place PR 1 fails, and for a reason that has
nothing to do with this PR. Observed on PR 1's run: all three fail at the step **`Setup Yarn 4.6`** with

```
```

Note *which* step that is: `threeal/setup-yarn-action@v2` performs the dependency install itself as part
of its caching, so the workflow's own, more obviously named `Install all dependencies` step is never
reached and is reported `skipped`. The cause is that the root `package.json` still declares `frontend`,
`docs` and `backend/vaa-strapi` as workspaces, while PR 1 moved two of them under `apps/` and PR 2
deleted the third. Those 412 packages dropping out of the resolution are the dependency closure of the
moved-away workspaces. The root `package.json` and `yarn.lock` are corrected in

**The green signal for this PR is therefore a local run, and it is a real one.** Against a database
reset from migrations — `yarn db:reset` applying 00001, 00002 and 00003 plus `seed.sql` — the pgTAP
suite reports:

```
Files=11, Tests=272, All tests successful, Result: PASS
```

It did not start there. See below.

Every item of `.agents/code-review-checklist.md` was swept against this slice and given a recorded
verdict. This is the **only** slice in the stack where the Supabase Backend block (9 items) and the Edge
Functions block (3 items) apply at all, so 24 cells were filled here against 12 for a typical slice. The
security review is **exhaustive, not sampled**: all 97 policies, all 9 `SECURITY DEFINER` functions, all
5 Edge Function files and all 3 migrations were read.

Where an item is enumerable it is reported as a count with the command that produced it, not as an
impression:

| Check | Checked | Conforming |
| --- | ---: | ---: |
| auth-identity calls in the `(SELECT auth.…())` scalar-subquery form | 56 occurrences | **56** — zero bare survivors |
| RLS policies naming an explicit `TO` role | 97 policies | **97** (`authenticated` 80, `anon` 15, `supabase_auth_admin` 1, `service_role` 1) |
| tables with RLS enabled | 20 tables | **19** — the 20th is in an unexposed schema and is correct by design |
| content tables carrying the standard five-policy set | 13 | **13** |
| `SECURITY DEFINER` functions pinning `SET search_path = ''` | 9 | **9** |
| FK columns with a covering index | 32 | **30** — the 2 gaps are named below |
| pgTAP files on the transaction-boundary pattern with `create_test_data()` | 11 | **11** |
| admin-gated Edge Functions verifying the caller before constructing a `service_role` client | 2 | **2** |

The order in that last row is the point: in both `invite-candidate` and `send-email` the caller check
*completes* before the privileged client is constructed, so there is no window in which a `service_role`
client exists for an unverified caller.

**Six findings were fixed on the development branch *before* this slice was cut**, so what you are
reading is the corrected code and this PR contains no fix of itself. Three of them are worth naming:

- **The `send-email` Edge Function had never worked.** It called the `resolve_email_variables` RPC with
  the argument names `user_ids` / `template_body` / `template_subject`; the SQL function declares
  `p_user_ids` / `p_template_body` / `p_template_subject`. PostgREST resolves overloads *by argument
  name*, so every invocation returned `PGRST202` and the handler faithfully converted it to HTTP 500.
  Verified against the running instance in both directions rather than argued from the source: the
  un-prefixed POST returns PostgREST's own hint naming the correct parameters, and the `p_`-prefixed
  POST returns `[]`. The error handling around it was well built and permanently taken.
- **The pgTAP suite was red, 2 of 272, and the mandated test pattern was the cause.** Migration 00002
  tightened `anon_select_candidates` to require `terms_of_use_accepted < now()`, and `create_test_data()`
  inserted the published fixture candidates with `ToU = now()` specifically to keep those assertions
  passing. But pgTAP wraps each file in one transaction — the `BEGIN`/`ROLLBACK` boundary the checklist
  itself requires — and `now()` is `transaction_timestamp()`, frozen for the transaction. So the guard
  evaluated `now() < now()`, which is false. Measured directly: `BEGIN; SELECT now() < now();` → `f`.
  Fixed in the fixture; no schema change.
- **`scripts/lint-schema.mjs` had never run against this project's database.** Its fallback URL named
  port `54332`; `config.toml` declares `54322`. A digit transposition. On the sweeping machine another
  Supabase instance answered on 54332, so the script silently linted *that* database and reported 149
  warnings about tables that do not exist in this repository. Pointed at the right database it then
  reported 51 warnings, because its foreign-key check compared `pg_constraint.conkey`, a 1-based
  `smallint[]`, against `pg_index.indkey`, an `int2vector`, **which is 0-based** — so every single-column
  foreign key was reported unindexed, including ones whose index demonstrably exists. Both defects are
  fixed; the corrected gate reports `0 errors, 2 warnings`, and those two rows match an independent
  static read of `200-indexes.sql` against the declared foreign keys.

**Two findings are escalated rather than deferred quietly**, and a reviewer should know they are open:

- **`yarn db:lint:sql` still exits non-zero**, on four `plpgsql_check` warnings. Two of them are unused
  parameters on `resolve_email_variables`, and the only change that turns the gate green is dropping
  them — a breaking signature change to a public RPC that is granted to two roles, present in the
  generated types, and called by a pgTAP test. Whether per-template variable resolution is still
  intended is a product question, not a cleanup, so it is left open deliberately.
- **The Signicat identity path keys account identity on `birthdate`.** `claimConfig.ts` sets
  `identityMatchProp: 'birthdate'`, and that value is used both to find an existing auth user and to
  derive the account's placeholder email — so two candidates sharing a birth date resolve to the same
  auth user and the same candidate record. `IDENTITY_PROVIDER_TYPE` defaults to `signicat`. This is
  **not** an Edge Function slip: the frontend states the same design independently, so changing this half
  alone would desynchronise the two. It is being decided across both halves at once. (Idura, the newer
  provider, correctly uses `sub`.)

**Three further items are recorded as knowingly shipping unfixed**: two join-table foreign keys lack a
covering index (the trailing columns of composite primary keys, so the reverse lookup and the
`ON DELETE CASCADE` are sequential scans); 22 of 52 triggers use naming prefixes the checklist does not
list, consistently but undocumented; and `public.storage_config` stores a service-role key in a plaintext
column, which is fail-closed today (RLS on, zero policies, `REVOKE ALL … FROM anon, authenticated,
public`) but whose production remedy is Supabase Vault.

**One item is deferred to a later PR in this stack**: the Developers' Guide still documents the Strapi
backend these 119 files replace, with no Supabase equivalent page. Those pages live under `apps/docs/**`
and are reviewed in PR 10.

Full evidence with file-and-line citations:
**"Slice 03 — `ship/v0.2-akita-03-supabase`"**.

[151-DISPOSITION.md § Slice 03](https://github.com/OpenVAA/voting-advice-application/blob/ship/v0.2-akita-11-planning/.planning/phases/151-ship-v0-2-akita-review-stack/151-DISPOSITION.md#slice-03--shipv02-akita-03-supabase)

119 files and 16,422 changed lines are both inside GitHub's rendering limits, so this PR displays in
full with no "Load more". Note that 62 of the files are benchmark harnesses and results; the reviewable
surface is closer to 57 files.

## The stack

`main`.

| **4** | **`ship/v0.2-akita-03-supabase`** ← *you are here* | replace the Strapi backend with the Supabase schema, RLS, functions and generated types | **119** | **16422** | **0** |
| 6 | `ship/v0.2-akita-05-e2e-tests` | add the Playwright end-to-end suite and its runner configuration | 195 | 23297 | 778 |
| 12 | `ship/v0.2-akita-11-planning` | add the v0.2 planning record and agent configuration | 2288+ | 866928 | 100 |

is what a reviewer actually sees. Rows below the current frontier are from the partition's dry run and
move slightly as fixes land on the development branch before each slice is cut.


apps/supabase/benchmarks/README.md
kaljarv, line 1: Move the results from the benchmarks in a concise format to supabase README and archive the benchmark scripts. You can point to a commit in the history where they'll be available if needed in the future.

apps/supabase/scripts/lint-schema.mjs
kaljarv, line 1: Check if we can write these as pgTap tests instead.

apps/supabase/supabase/functions/identity-callback/index.ts
```diff
@@ -0,0 +1,397 @@
...
+  if (req.method !== 'POST') {
+    return new Response(JSON.stringify({ error: 'Method not allowed' }), {
+      status: 405,
+      headers: { ...corsHeaders, 'Content-Type': 'application/json' }
+    });
+  }
+
+  try {
+    // Resolve provider configuration
+    const providerType = Deno.env.get('IDENTITY_PROVIDER_TYPE') ?? 'signicat';
```
kaljarv, line 168: Remove default provider

apps/supabase/supabase/functions/identity-callback/index.ts
```diff
@@ -0,0 +1,397 @@
...
+    const { firstName, lastName, matchValue: identityMatchValue, extraClaims: extractedClaims } = claimResult;
+
+    // 5. Create admin Supabase client for user/candidate operations
+    const supabaseAdmin = createClient(Deno.env.get('SUPABASE_URL')!, Deno.env.get('SUPABASE_SERVICE_ROLE_KEY')!);
+
+    // 6. Find or create auth user by identity_match_value
+    let userId = await findUserByIdentityMatch(supabaseAdmin, identityMatchValue);
+    let isNewUser = false;
+
+    // Bank-auth identities carry no real email at sign-in (the candidate is prompted to
```
kaljarv, line 255: Rem line breaks from comments

apps/supabase/supabase/functions/identity-callback/index.ts
```diff
@@ -0,0 +1,397 @@
...
+    // Use generateLink with magiclink type to create a login URL, addressed to the same
+    // identity-derived placeholder email the user record was created with above so the
+    // admin generateLink API resolves to the existing user.
+    const siteUrl = Deno.env.get('SUPABASE_URL')!.replace(/\/+$/, '');
+
+    const { data: linkData, error: linkError } = await supabaseAdmin.auth.admin.generateLink({
+      type: 'magiclink',
+      email: placeholderEmail,
+      options: {
+        redirectTo: `${Deno.env.get('SITE_URL') || 'http://127.0.0.1:5173'}/candidate`
```
kaljarv, line 341: Remove defaults or construct from other env vars fully. Search the repo for any hard-coded ports or localhost urls.

apps/supabase/supabase/functions/identity-callback/index.ts
```diff
@@ -0,0 +1,397 @@
...
+    const { id_token, project_id } = body;
+
+    if (!id_token || typeof id_token !== 'string') {
+      return new Response(JSON.stringify({ error: 'Missing or invalid id_token' }), {
+        status: 400,
+        headers: { ...corsHeaders, 'Content-Type': 'application/json' }
+      });
+    }
+
+    const projectId = project_id || Deno.env.get('DEFAULT_PROJECT_ID') || DEFAULT_SEED_PROJECT_ID;
```
kaljarv, line 196: Remove this default and throw if the env is not available.

apps/supabase/supabase/functions/identity-callback/claimConfig.ts
```diff
@@ -0,0 +1,87 @@
...
+ * Per-provider claim configurations.
+ *
+ * - Signicat: Uses `birthdate` for identity matching (Finnish bank auth pattern).
+ *   No extra claims extracted.
+ * - Idura: Uses `sub` (persistent pseudonym) for identity matching.
+ *   Extracts `birthdate` and `hetu` (Finnish personal identity code) as metadata.
+ */
+export const PROVIDER_CONFIGS: Record<string, ProviderClaimConfig> = {
+  signicat: {
+    identityMatchProp: 'birthdate',
```
kaljarv, line 35: Check the current Signicat docs whether there's a pseudonym we can use for identity matching and if not use hetu which should be available. If not, remove Signicat support completely bc b-day-based matching is not unique.

apps/supabase/supabase/functions/send-email/index.ts
```diff
@@ -0,0 +1,300 @@
...
+        headers: { ...corsHeaders, 'Content-Type': 'application/json' }
+      });
+    }
+
+    // -------------------------------------------------------------------------
+    // 3. Create admin client (service_role) and resolve template variables
+    // -------------------------------------------------------------------------
+    const supabaseAdmin = createClient(Deno.env.get('SUPABASE_URL')!, Deno.env.get('SUPABASE_SERVICE_ROLE_KEY')!);
+
+    // Argument names must match the SQL function's parameter names exactly:
```
kaljarv, line 130: Rem line breaks from comments

apps/supabase/supabase/functions/send-email/index.ts
```diff
@@ -0,0 +1,300 @@
...
+
+    for (const recipient of recipients) {
+      // Select template matching recipient's preferred locale, fallback to first available
+      const locale = availableLocales.includes(recipient.preferred_locale) ? recipient.preferred_locale : defaultLocale;
+
+      const template = templates[locale];
+      const vars: Record<string, string> = recipient.variables || {};
+
+      // Replace {{variable.path}} placeholders with resolved values
+      const replaceVars = (text: string): string =>
```
kaljarv, line 175: Allow spaces like `{{ varname }}`

apps/supabase/supabase/functions/send-email/index.ts
```diff
@@ -0,0 +1,300 @@
...
+          results: dryResults
+        }),
+        { status: 200, headers: { ...corsHeaders, 'Content-Type': 'application/json' } }
+      );
+    }
+
+    // -------------------------------------------------------------------------
+    // 6. Send emails via SMTP
+    // -------------------------------------------------------------------------
+    const smtpHost = Deno.env.get('SMTP_HOST') || 'inbucket';
```
kaljarv, line 210: Remove all defaults and throw if not available, also below.

apps/supabase/supabase/schema/000-enums.sql
```diff
@@ -0,0 +1,26 @@
...
+CREATE TYPE public.entity_type AS ENUM (
+    'candidate', 'organization', 'faction', 'alliance'
+);
+
+CREATE TYPE public.category_type AS ENUM (
+    'info', 'opinion', 'default'
+);
+
+CREATE TYPE public.user_role_type AS ENUM (
+    'candidate', 'party', 'project_admin', 'account_admin', 'super_admin'
```
kaljarv, line 25: Rename party to organization. When making schema changes, we can rewrite both migrations and schemata and not care about bwd compat. Make sure to check db seed scripts as well.

apps/supabase/supabase/schema/011-validation-functions.sql
```diff
@@ -0,0 +1,298 @@
...
+      END IF;
+      -- Each array item must be a string or localized string
+      FOR p_item IN SELECT * FROM jsonb_array_elements(p_answer_value)
+      LOOP
+        IF jsonb_typeof(p_item) != 'string' AND NOT public.is_localized_string(p_item) THEN
+          RAISE EXCEPTION 'Each item in multipleText answer must be a string or localized string object';
+        END IF;
+      END LOOP;
+    WHEN 'image' THEN
+      IF jsonb_typeof(p_answer_value) != 'object' THEN
```
kaljarv, line 165: Extract the full check into a is_image util.

apps/supabase/supabase/schema/000-enums.sql
```diff
@@ -0,0 +1,26 @@
...
+
+CREATE TYPE public.question_type AS ENUM (
+    'text', 'number', 'boolean', 'image', 'date', 'multipleText',
+    'singleChoiceOrdinal',
+    'singleChoiceCategorical',
+    'multipleChoiceCategorical'
+);
+
+CREATE TYPE public.entity_type AS ENUM (
+    'candidate', 'organization', 'faction', 'alliance'
```
kaljarv, line 17: Slightly unrelated, check the whole repo's symbol naming for whether it follows UK or US spelling. If there are any UK variants, change them to US ones for consistency.

apps/supabase/supabase/schema/103-questions.sql
```diff
@@ -0,0 +1,114 @@
...
+  color           jsonb,
+  image           jsonb,
+  sort_order      integer,
+  subtype         text,
+  custom_data     jsonb,
+  is_generated    boolean       DEFAULT false,
+  created_at      timestamptz   NOT NULL DEFAULT now(),
+  updated_at      timestamptz   NOT NULL DEFAULT now(),
+  category_type   public.category_type DEFAULT 'opinion',
+  election_ids    jsonb,
```
kaljarv, line 20: Evaluate whether stricter fkey linkage would be needed for these id-jsonbs, and also in questions. The challenge is that we'll want to extend the logic to support to three options (but not a combination of these):
1. only included in elections with these ids
2. excluded from elections with these ids
3. no filter

apps/supabase/supabase/schema/104-nominations.sql
```diff
@@ -0,0 +1,61 @@
...
+      WHEN candidate_id IS NOT NULL THEN 'candidate'::public.entity_type
+      WHEN organization_id IS NOT NULL THEN 'organization'::public.entity_type
+      WHEN faction_id IS NOT NULL THEN 'faction'::public.entity_type
+      WHEN alliance_id IS NOT NULL THEN 'alliance'::public.entity_type
+    END
+  ) STORED,
+  -- Election context
+  election_id          uuid        NOT NULL REFERENCES public.elections(id) ON DELETE CASCADE,
+  constituency_id      uuid        NOT NULL REFERENCES public.constituencies(id) ON DELETE CASCADE,
+  election_round       integer     DEFAULT 1,
```
kaljarv, line 46: Add a constraint >= 1

apps/supabase/supabase/schema/107-feedback.sql
```diff
@@ -0,0 +1,94 @@
...
+-- RLS policies are in 302-rls.sql.
+-- Rate limiting trigger prevents spam (5 requests per 5 minutes per IP).
+
+--------------------------------------------------------------------------------
+-- Private schema for rate limiting (not exposed via PostgREST)
+--------------------------------------------------------------------------------
+CREATE SCHEMA IF NOT EXISTS private;
+
+CREATE TABLE IF NOT EXISTS private.feedback_rate_limits (
+  ip_address   text        PRIMARY KEY,
```
kaljarv, line 14: Investigate in a follow-up task whether there is a way to encrypt the ip address so that it's not recoverable afterwards or use some other non-PII moniker.

apps/supabase/supabase/schema/300-auth-tables.sql
```diff
@@ -0,0 +1,67 @@
...
+--             All voter-facing content tables
+
+--------------------------------------------------------------------------------
+-- user_roles table
+--------------------------------------------------------------------------------
+CREATE TABLE public.user_roles (
+  id         uuid           PRIMARY KEY DEFAULT gen_random_uuid(),
+  user_id    uuid           NOT NULL REFERENCES auth.users(id) ON DELETE CASCADE,
+  role       user_role_type NOT NULL,
+  scope_type text           NOT NULL,  -- 'candidate', 'party', 'project', 'account', 'global'
```
kaljarv, line 14: This should use an enum matching user_role_type prefixes. However, we'll want to refactor the whole auth model a bit. We'll have to do that in a follow-up phase (blocking ship).

apps/supabase/supabase/schema/301-auth-functions.sql
```diff
@@ -0,0 +1,150 @@
...
+-- Usage in RLS policies:
+--   has_role('super_admin')                             -- any super_admin role
+--   has_role('project_admin', 'project', project_id)    -- project_admin for specific project
+--   has_role('candidate', 'candidate', candidate_id)    -- candidate for specific record
+--
+-- SECURITY DEFINER with empty search_path to prevent search_path attacks.
+-- Uses (SELECT auth.jwt()) to ensure a single evaluation per query.
+--------------------------------------------------------------------------------
+CREATE OR REPLACE FUNCTION public.has_role(
+  p_check_role text,
```
kaljarv, line 59: Use enums

apps/supabase/supabase/schema/302-rls.sql
```diff
@@ -0,0 +1,497 @@
...
+CREATE POLICY "anon_select_elections" ON public.elections FOR SELECT TO anon
+  USING (published = true);
+
+CREATE POLICY "authenticated_select_elections" ON public.elections FOR SELECT TO authenticated
+  USING (
+    (SELECT can_access_project(project_id))
+    OR published = true
+  );
+
+CREATE POLICY "admin_insert_elections" ON public.elections FOR INSERT TO authenticated
```
kaljarv, line 88: We need to separate can_access_project from can_edit_project. For example, all candidates should be able to read the project's data like elections but never edit them.

apps/supabase/supabase/schema/102-entities.sql
```diff
@@ -0,0 +1,90 @@
...
+);
+
+CREATE TRIGGER set_updated_at
+  BEFORE UPDATE ON public.organizations
+  FOR EACH ROW EXECUTE FUNCTION public.update_updated_at();
+
+CREATE TABLE public.candidates (
+  id              uuid        PRIMARY KEY DEFAULT gen_random_uuid(),
+  project_id      uuid        NOT NULL REFERENCES public.projects(id) ON DELETE CASCADE,
+  name            jsonb,
```
kaljarv, line 27: This probably conflicts with first and last name based names, so should be removed. Short_name, tho, is still valid as an override for generated initials.

apps/supabase/supabase/schema/303-column-grants.sql
```diff
@@ -0,0 +1,56 @@
...
+-- Protected (admin-only) columns:
+--   project_id      - determines project tenancy
+--   auth_user_id    - links candidate to auth user, set during invite/registration
+--   organization_id - party assignment
+--   published       - publication status, admin-controlled
+--   id              - primary key, immutable
+--   is_generated    - system flag for mock/generated data
+--
+-- Allowed columns for candidates (self-edit):
+--   name, short_name, info, color, image, sort_order, subtype,
```
kaljarv, line 28: sort_order is should not be self-editable.

apps/supabase/supabase/schema/503-entity-rpcs.sql
```diff
@@ -0,0 +1,189 @@
...
+
+GRANT EXECUTE ON FUNCTION public.get_candidate_user_data(public.entity_type) TO authenticated;
+
+--------------------------------------------------------------------------------
+-- upsert_answers: atomic answer write for a single entity
+--
+-- SECURITY INVOKER: runs with caller's permissions, so RLS policies enforce
+-- that a candidate can only update their own row.
+--------------------------------------------------------------------------------
+CREATE OR REPLACE FUNCTION public.upsert_answers(
```
kaljarv, line 147: This should be extended to affect all entities with answers, i.e. organzations as well.

apps/supabase/supabase/schema/504-admin-rpcs.sql
```diff
@@ -0,0 +1,36 @@
...
+-- Functions:
+--   merge_custom_data() - shallow JSONB merge on questions.custom_data
+
+--------------------------------------------------------------------------------
+-- merge_custom_data: shallow JSONB merge on questions.custom_data
+--
+-- SECURITY INVOKER: the existing admin_update_questions RLS policy enforces
+-- that only admins with can_access_project() can update questions.
+--------------------------------------------------------------------------------
+CREATE OR REPLACE FUNCTION public.merge_custom_data(
```
kaljarv, line 12: Either rename to merge_question_custom_data or make generic to target other such tables as well.

apps/supabase/supabase/config.toml
```diff
@@ -0,0 +1,399 @@
...
+max_indexes = 5
+
+# Vector Buckets is available to Supabase Pro plan.
+# [storage.vector.buckets.documents-openai]
+
+[auth]
+enabled = true
+# The base URL of your website. Used as an allow-list for redirects and for constructing URLs used
+# in emails.
+site_url = "http://127.0.0.1:5173"
```
kaljarv, line 165: Investigate whether this and the other hard-coded ports can be gotten from env. Otherwise make sure that if the where and when the port is changed, this is mentioned as a caveat.

kaljarv (commented):
Checked.

PR #867, opened 2026-08-17
5/12 feat: add the dev-seed package that generates deterministic local and E2E data

kaljarv:
**Stack position: 5 of 12.** Base: `ship/v0.2-akita-03-supabase`. Head: `ship/v0.2-akita-04-dev-seed`.
**162 files changed, 19,661 insertions, 0 deletions** — all 162 are additions.

---

   zero files.

every off-diagonal cell of the pairwise overlap matrix is `0`, and the slices plus an empty remainder
reconstruct the target tree exactly. Re-verified after the slice above this one was cut: the six cut
slices plus the untouched remainder produce tree `8459312c9`, which is the development branch's tree
hash, and the partition arithmetic closes at a gap of `0` — 252 + 97 + 119 + 162 + 195 + 3,458 = 4,283,
the comparable file total.)

## What you are looking at

`@openvaa/dev-seed` — the package that generates the local development dataset and every dataset the
E2E suite runs against. It replaces the Strapi-era `GENERATE_MOCK_DATA_ON_RESTART` bootstrap that PR 2
deleted, and it writes into the Supabase schema PR 4 added, through that schema's `bulk_import` RPC
with a service-role client.

| Area | Files |
| --- | ---: |
| `packages/dev-seed/src/**` — CLIs, generators, the latent answer model, template registry | 113 |
| `packages/dev-seed/tests/**` — unit + integration tests | 43 |
| `packages/dev-seed/{README.md, package.json, tsconfig.json, vitest.config.ts, scripts/}` | 5 |
| `apps/frontend/static/images/e2e-test-image-1.jpg` — the portrait the E2E seed template uploads | 1 |

By extension: **127 `.ts`**, **31 `.jpg`** (30 committed portraits plus the E2E fixture image), 2 `.md`,
2 `.json`.

- **The two CLIs are the whole user-facing surface**: `yarn db:seed` (`src/cli/seed.ts`) and
  `yarn db:seed:teardown` (`src/cli/teardown.ts`). Both run `parseArgs` with `strict: true` and
  `allowPositionals: false`, so an unknown flag is rejected rather than ignored, and `--seed` is
  validated against `/^-?\d+$/` on the whole token before `parseInt` — because `parseInt('12abc')`
  returns `12`, and a typo'd seed that silently becomes a different-but-valid dataset is the worst
  failure mode a deterministic generator has.
- **Determinism is the point.** A given `--seed` plus `--external-id-prefix` produces the same dataset
  every time, which is what lets the E2E suite assert exact counts and exact match scores rather than
  ranges. The answer model is a latent-factor emitter (`src/latent/**`), not per-row randomness.
- **`src/templates/` is 33 of the 113 `src` files and holds 30 built-in templates** — `default` (the
  Finnish demo, 4 locales), `e2e/base` (the canonical E2E dataset), and 28 `e2e/perm/*` permutation
  overlays, one per app-settings configuration the suite exercises. 17 of the 30 compose the shared
  `_helpers/buildMinimal.ts`; the 13 that do not are individually justified at
  `src/templates/index.ts:104-124` as layouts the helper cannot express.
- **`seed:teardown` deletes rows.** It is prefix-scoped and the allowed table list is a single exported
  constant (`src/cli/teardown.ts:73-85`) that the tests import rather than mirror — but read the
  locality note under "Two things a reviewer should know are open" below before running it.

## Why no checks run here

`ship/v0.2-akita-03-supabase`, a sibling branch, so **no checks run — and that is correct behaviour,
not a misconfiguration.** Only PR 1 of the stack fires CI at all.

**Do not read that as "a job exists and was skipped."** The workflow file at this branch's head is
`frontend-and-shared-module-validation`, `backend-validation` and `e2e-tests`. **The
`dev-seed-integration` job that exercises this package does not exist at this head.** It arrives with
PR 11, where this branch's own updated workflow file lands. (For the record, because it is easy to
assume the opposite: that job deliberately carries **no** paths filter — `main.yaml:130-136` in the
newer file names the incident that made it unconditional. It is not conditional; it is simply not
present yet.)

Had the three jobs that *are* present run, they would have failed at the same place PR 1 fails, and for
a reason that has nothing to do with this PR. Observed on PR 1's run: all three fail at the step
**`Setup Yarn 4.6`** with

```
```

Note *which* step that is: `threeal/setup-yarn-action@v2` performs the dependency install itself as part
of its caching, so the workflow's own, more obviously named `Install all dependencies` step is never
reached and is reported `skipped`. The cause is that the root `package.json` still declares `frontend`,
`docs` and `backend/vaa-strapi` as workspaces, while PR 1 moved two of them under `apps/` and PR 2
deleted the third. Those 412 packages dropping out of the resolution are the dependency closure of the
moved-away workspaces. The root `package.json` and `yarn.lock` are corrected in

**The green signal for this PR is therefore a local run, and it is a real one.**

```
yarn workspace @openvaa/dev-seed test:unit
  Test Files  43 passed (43)
       Tests  446 passed (446)
```

That includes `tests/integration/default-template.integration.test.ts`, which writes the full `default`
template against a running local Supabase in 11.8 s and asserts an operation budget while doing it — so
the integration path is covered by an executing test, not only by unit mocks.

recorded verdict: 12 cells, all terminal. The three conditional blocks (Supabase Backend, Supabase
Adapter, Edge Functions) do not apply — this slice touches neither `apps/supabase/` nor the frontend
adapter — and two general items are recorded `n/a` with their reason: there is no `.svelte` file in the
slice, and nothing in it renders markup to a browser.

Where an item is enumerable it is reported as a count with the command behind it, not as an impression:

| Check | Checked | Result |
| --- | ---: | ---: |
| `any` in type position in `src/` | 116 TS files | **0** — every occurrence is in `tests/`, in mock-shape surfaces |
| credential-shaped literals across the slice | 162 files | **0** secrets — see the note on the two `eyJ…` literals below |
| `writeFile` / `unlink` / `rm` anywhere in `src/` | 113 files | **0** — a template-supplied string cannot reach the filesystem |
| `TODO` / `FIXME` / `HACK` / `XXX` | 162 files | **0** |
| `Foo[]` array-suffix type declarations (the style guide requires `Array<Foo>`) | 116 TS files | **0** — a bare grep returns one hit, inside a comment quoting the supabase-js API |
| built-in templates documented in the README | 30 in the registry | **30** after this sweep — it was **2**, and one of the two was a name that no longer exists |
| names enumerated in `src/index.ts`'s "Public API (stable)" docblock | 38 exported | **38** after this sweep — it was **18**, with 0 phantom entries |

**Six findings were fixed on the development branch *before* this slice was cut**, so what you are

- **The README documented a built-in template that no longer exists.** It named `e2e` at three sites,
  including a worked troubleshooting example; the registry has 30 keys and no `e2e`, so
  `--template e2e` errors out. The real name is `e2e/base`. The cross-check that settles which document
  was stale: the repository's root `CLAUDE.md` has it right and records the retirement explicitly — so
  the repo-level document was updated when the name changed and the package's own README was not.
- **`src/index.ts`'s "Public API (stable)" docblock enumerated 18 of the 38 names the file exports.**
  The 20 omissions were not marginal: the entire teardown surface, the entire built-in-template surface
  and the whole locale surface. Zero phantom entries, so the inventory was incomplete rather than wrong
  — which is the failure mode that survives review, because everything it says is true.
- **A `catch` discarded the one cause that mattered.** `cli/seed.ts`'s `loadBuiltIns()` caught *every*
  error from the dynamic import of the template registry and returned empty maps, so any runtime failure
  inside `src/templates/index.ts` reached the user as
  `Unknown template: 'default'. Built-in templates: (none registered yet).` — a message that names the
  wrong cause and sends the reader off to check their spelling. The cause is now reported on stderr; the
  fallback itself is unchanged.
- **Four `eslint-disable @typescript-eslint/no-explicit-any` suppressions carried no reason.** Each now
  states *why* the type cannot be written — an unexported deep supabase-js generic, a self-referential
  thenable assembled by assigning its own methods onto itself, and an array of ad-hoc literals that are
  not instances of the class they stand in for. Recorded alongside the fix, because it is the more useful
  fact: **the rule could not have flagged any of the four**, since the package's lint script is
  `src/`-only and all 43 `tests/` files are unlinted.

**On the two `eyJ…`-shaped literals** (`src/supabaseAdminClient.ts:42`,
`tests/integration/default-template.integration.test.ts:126`): both decode to
`{"iss":"supabase-demo","role":"service_role"}` — the published Supabase CLI local demo key, the same
value `apps/supabase/seed.sql` seeds. Not a secret, and named here so nobody has to decode them.

- **`seed:teardown` has no locality guard, and the README now says so instead of claiming one.** The
  documentation previously asserted "the writer refuses to run without `SUPABASE_URL` set" as the
  package's guard. That is true of `Writer` — and false of the destructive command, which never
  constructs a `Writer`: `seed:teardown` goes through `SupabaseAdminClient` directly, whose module-level
  fallbacks supply a URL and the demo key with no enforcement. Both CLIs additionally fall back
  `SUPABASE_URL ??= PUBLIC_SUPABASE_URL` — the *deployed frontend's* variable — so a repo-root `.env`
  configured for a non-local Supabase silently retargets both commands, including the mass-delete. **The
  documentation is fixed; the guard itself is left to a deliberate decision**, because adding one changes
  the behaviour of a command that deletes rows and CI legitimately points these CLIs at a non-`localhost`
  test instance. Deletes are prefix-scoped and the prefix guard rejects prefixes under two characters, so
  the blast radius is bounded even in that case — but it is bounded by the prefix, not by the host.
- **This package contributes 15 of the repository's 20 `lint:check` warnings, and they are staying.**
  All 15 are one deliberate class: `'ctx' is defined but never used` on the uniform
  `generate(fragment, ctx)` generator signature, which every generator implements whether or not it needs
  the context, plus one deliberately discarded `external_id`. The rule's own remedy is an underscore
  prefix, and applying it would take the repository-wide gate from `0 errors / 20 warnings` to `0 / 5` —
  moving a number that this review stack's own verification compares "unchanged" against, in eight
  separate places. Re-baselining a shared gate mid-stack to make one slice look tidier is worse than
  recording the gap, so it is recorded.

**One documentation fix for this package landed in a different PR, and it is worth knowing where.** The
Developers' Guide's testing page made three false claims about this package — the template name, a
`variant-*.ts` path that matches zero tracked files, and "seeded via the Admin Tools API" (it is the
`bulk_import` RPC; Admin Tools was the Strapi plugin PR 2 deleted). Those pages live under `apps/docs/**`
and are reviewed in **PR 10**.

162 files and 19,661 changed lines are both inside GitHub's rendering limits, so this PR displays in full
with no "Load more" — but the line count is only **339 under the 20,000 cap**, so it should not be added
to casually. 31 of the 162 files are binary `.jpg` portraits and render as image diffs; the reviewable
text surface is 131 files.

## The stack

`main`.

| **5** | **`ship/v0.2-akita-04-dev-seed`** ← *you are here* | add the dev-seed package that generates deterministic local and E2E data | **162** | **19661** | **0** |
| 12 | `ship/v0.2-akita-11-planning` | add the v0.2 planning record and agent configuration | 2314 | 866928 | 100 |

Line counts are measured over each slice's own commit. Row 1's is the rename-aware measurement, which is
what a reviewer actually sees. Rows below the current frontier are from the partition's dry run and move
slightly as fixes land on the development branch before each slice is cut.

## Full evidence

**"Slice 04 — `ship/v0.2-akita-04-dev-seed`"**.

[151-DISPOSITION.md § Slice 04](https://github.com/OpenVAA/voting-advice-application/blob/ship/v0.2-akita-11-planning/.planning/phases/151-ship-v0-2-akita-review-stack/151-DISPOSITION.md#slice-04--shipv02-akita-04-dev-seed)


kaljarv (commented):
Passed as non-essential.

PR #868, opened 2026-08-17
6/12 test: add the Playwright end-to-end suite and its runner configuration

kaljarv:
# 6/12 · `test: add the Playwright end-to-end suite and its runner configuration`

**Stack position: 6 of 12.** Base: `ship/v0.2-akita-04-dev-seed`. Head: `ship/v0.2-akita-05-e2e-tests`.
**195 files changed, 23,325 insertions, 778 deletions.**

**This PR is the whole test tree, as one commit.** That is deliberate and it satisfies a stated
requirement of the restructure: *all tests are one commit*. `git log --oneline base..head` returns
exactly **1**, and there is no second test-typed commit anywhere in the stack to collapse into it.
The coincidence is structural rather than lucky — this slice's path filter is the single token
`tests`, so "one slice", "one commit" and "all tests" are the same set by construction.

---

   zero files.

every off-diagonal cell of the pairwise overlap matrix is `0`. Re-verified after the slice above this
one was cut: the seven cut slices plus the untouched remainder produce tree `06f7ab4f4`, which is the
development branch's tree hash, and the partition arithmetic closes at a gap of `0` —
252 + 97 + 119 + 162 + 195 + 533 + 2,934 = 4,292, the comparable file total. The remainder also
decomposes independently, without reference to the catch-all: 213 + 329 + 39 + 37 + 2,316 = 2,934.)

## What you are looking at

The root `tests/` tree: the Playwright end-to-end suite, its fixtures, its runner configuration, and
its support harness. It runs against the frontend the next two PRs rewrite, seeded by the
`@openvaa/dev-seed` package PR 5 added, through the Supabase schema PR 4 added.

| Area | Files |
| --- | ---: |
| `tests/tests/specs/**` — the specs themselves | 43 added `.spec.ts` (plus 3 deleted) |
| `tests/tests/fixtures/**`, `tests/tests/support/**`, `tests/tests/utils/**` — page fixtures, the mock OIDC issuer, test ids, scans | ~120 |
| `tests/playwright.config.ts`, `tests/README.md`, runner + type config | ~10 |
| deleted Strapi-era specs and helpers | 7 |

By extension: **171 `.ts`** in the diff; 191 tracked `.ts`/`.md`/config files under `tests/` at this head.

- **`tests/playwright.config.ts` is where the suite's shape lives** — the project graph, the
  dependency chain, and three guards that run at *config-load* time and abort the whole run rather
  than fail a spec: an orphan-probe check, a soft-assertion budget, and a teardown-prefix uniqueness
  check. `npx playwright test --list` exercises all three; it reports **143 tests in 94 files** and
  passes.
- **The suite has a served-application preflight** in global setup that cannot be skipped by any flag
  or environment variable. It asserts that the page under test came from *this* checkout — the server
  must serve and echo back this working tree's absolute path — not merely that something answered on
  the port. `tests/README.md` § Run explains how to read a preflight failure field by field.
- **`tests/support/mock-oidc-*.pem` commits a private key, and that is intentional and documented.**
  It is a 2048-bit RSA test key with `CN=127.0.0.1`, un-ignored by an explicit two-line `.gitignore`
  exception carrying a four-line rationale; the issuer binds `127.0.0.1` only, never a public
  interface. `tests/utils/testKeys.ts` does the same for its inline JWKs and states the failure mode
  outright ("reusing these in prod would enable id_token spoofing"). Named here so nobody has to
  wonder.
- **`tests/.gitignore` is load-bearing** and is one of three files in this tree that appear in *no*
  slice's diff (byte-identical across the layout move). Its `playwright*/` rule is what keeps the
  visual chain's `storageState` — a live candidate session cookie — out of git. Asserted:
  `git ls-files tests/` matches `playwright/`, `.auth` and `blob-report` **0** times each.

## Why no checks run here

`ship/v0.2-akita-04-dev-seed`, a sibling branch, so **no checks run — and that is correct behaviour,
not a misconfiguration.** Only PR 1 of the stack fires CI at all. Asserted, not predicted:
`gh pr checks` on the four already-open stacked PRs returns *"no checks reported"* on each.

**Do not read that as "the e2e job exists and was skipped."** The workflow file at this branch's head
is still `main`'s — byte-identical, blob `c2fdcedb2` — and it defines exactly three jobs:

**`Setup Yarn 4.6`** with

```
```

**Being honest about the green signal for THIS PR, because it is the one PR where it matters most.**
Every other PR in this stack can point at a local run. This one cannot yet, and saying otherwise would
undercut the suite's whole purpose:

> **The 43 specs in this PR have not been run.** Running them needs a dev server on `:5173` and a
> seeded local Supabase, and this project's own rule is that a did-not-run E2E test counts as a
> failure rather than a pass. **So this PR does not claim a green suite.** The suite is run once, in
> full, against the post-review-sweep tip of the development branch, and that run — not per-PR CI and
> not this PR — is the stack's collective proof. It is the last step of the work that produced this
> stack.

What *was* obtained is real and has a real failure mode: `npx playwright test --list` loads
`playwright.config.ts` with all three config-load guards active and reports 143 tests in 94 files.
A dropped spec, an orphaned probe or a drifted soft-assertion budget would have failed it. It is not
a suite pass and is not offered as one. `yarn typecheck:tests` is clean.

recorded verdict: 12 cells, all terminal. The three conditional blocks (Supabase Backend, Supabase
Adapter, Edge Functions) do not apply. Two general items are recorded `n/a` with their reason: there
is no `.svelte` component in the slice, and nothing in it defines user-facing app functionality that
would need a tracking event.

Where an item is enumerable it is reported as a count with the command behind it, not as an impression:

| Check | Checked | Result |
| --- | ---: | ---: |
| `any` in type position | 195 files | **0** — the cleanest result in the stack on this axis |
| secret-shaped literals (`eyJ…`, `password/secret/api_key/service_role` assignments) | 195 files | **0** — cleaner than PR 5, which carries two published-demo-key literals |
| `ignoreHTTPSErrors` | whole config | **1**, on the opt-in `bank-auth-journey` project only, never on the top-level `use`, with 11 lines saying why |
| `test.skip` / `test.fixme` / `test.only` / bare `skip(` / `fixme(` | 43 specs | **0**, before and after the sweep |
| visual baselines regenerated | this slice's commits | **0** — `-snapshots` and `__screenshots__` match 0 times |
| tracked files under `playwright/`, `.auth`, `blob-report` | `git ls-files tests/` | **0** each |
| specs the documented "full suite" actually runs | 43 | **38 before this sweep** — see below |

**52 defects were fixed on the development branch *before* this slice was cut**, so what you are

- **The README described a "full suite" that skipped 5 of the 43 specs.** The project inventory omitted
  the `_probes` project entirely, while the `test:e2e` script appends `--grep-invert @probe` — so 12%
  of the suite never ran in what the documentation called a full run, and the script that *does* run
  them was undocumented. Both fixed.
- **Three files declared a rigidity contract they violated.** `candidate-journey.spec.ts` states
  "0 `expect.soft`" and carries 3; two page fixtures state "NO `expect.soft`" and carry 4 and 6. In the
  one slice whose credibility every other slice borrows, a file describing its own rigour falsely is
  the finding. Each header was made **true** — the count, where the soft calls are, why soft is right
  there, and that the file sits outside the config's soft-assertion budget table. The budget table was
  deliberately *not* widened; the config reserves that for an explicit decision and now names these
  three files as a recorded follow-up.
- **38 comments had been left ungrammatical by an earlier automated cleanup pass, and one was
  load-bearing** — the sentence recording a knowingly-deferred decision about a config guard's scope,
  rendered unreadable. All 38 repaired. Found by comparing the cleanup commit's own before/after line
  pairs rather than by grepping the tree; a grep for the same damage signatures returns 672 candidates,
  almost all of them ordinary `...` spread syntax and possessives.
- **The Developers' Guide made six false claims about this suite** — a dataset "imported via Admin
  Tools API" (it is a `bulk_import` RPC; Admin Tools was the Strapi plugin PR 2 deletes), an auth step
  described as part of the default chain when it is declared only under a visual-testing flag, two
  environment variables with zero code references, and a `yarn dev:down` command that does not exist.
  Those pages live under `apps/docs/**` and are reviewed in **PR 10**.

- **Keyboard navigation has no automated gate anywhere in this suite, and the accessibility gate that
  does exist reaches 5 URLs.** The axe scan is default-on and asserts against all four WCAG tags, but
  its route table resolves to **5 distinct URLs × 2 themes** — measured, not estimated — which leaves
  **31 of the 36 page routes never scanned**, including every candidate-app route, every admin route,
  and the voter question flow. The raw-i18n-key scan has exactly the same reach, because its only call
  site is inside the axe helper. And axe is a static-DOM auditor: it does not tab through the UI, so
  the keyboard half of the requirement is uncovered even on the 5 URLs it does reach. Widening the
  route table and adding a keyboard gate is net-new test authorship and is scoped to a later phase.
- **CI sets `retries: 3`, which can turn a flake green.** Accepted, because the remedy already exists
  and is in this PR: `determinism-batch.sh` **refuses to start** with `CI` set, for exactly this
  reason, and fails a batch outright when the flaky count is non-zero; four projects pin `retries: 0`
  where a retry would mask a state-mutating assertion. The full-suite run that is this stack's real
  gate is local and non-CI at `retries: 0`.

195 files is inside GitHub's file budget, but **23,325 + 778 = 24,103 changed lines is over the
20,000-line diff cap**, so the Files-changed tab will collapse part of this PR behind "Load diff".
That is known and accepted rather than fixed by re-partitioning: splitting the test tree would break
the *all tests are one commit* requirement this PR exists to satisfy. If you want a guided path, the
per-item verdicts with file-and-line citations are linked at the bottom.

## The stack

`main`.

| **6** | **`ship/v0.2-akita-05-e2e-tests`** ← *you are here* | add the Playwright end-to-end suite and its runner configuration | **195** | **23325** | **778** |
| 12 | `ship/v0.2-akita-11-planning` | add the v0.2 planning record and agent configuration | 2316 | 866928 | 100 |

is what a reviewer actually sees. Rows below the current frontier are from the partition's dry run and
move slightly as fixes land on the development branch before each slice is cut — row 7 is measured
(it was cut alongside this PR's opening) and rows 8–12 are not yet.

## Full evidence

**"Slice 05 — `ship/v0.2-akita-05-e2e-tests`"**.

[151-DISPOSITION.md § Slice 05](https://github.com/OpenVAA/voting-advice-application/blob/ship/v0.2-akita-11-planning/.planning/phases/151-ship-v0-2-akita-review-stack/151-DISPOSITION.md#slice-05--shipv02-akita-05-e2e-tests)


kaljarv (commented):
Cursorily checked as non-essential.

PR #869, opened 2026-08-17
7/12 feat: rewrite the frontend library layer on Svelte 5 runes and the Supabase adapter

kaljarv:
# 7/12 · `feat: rewrite the frontend library layer on Svelte 5 runes and the Supabase adapter`

**Stack position: 7 of 12.** Base: `ship/v0.2-akita-05-e2e-tests`. Head: `ship/v0.2-akita-06-frontend-lib`.
**GitHub will show you 528 files changed, 22,550 insertions, 8,179 deletions.** Measured without
rename detection — the convention used for every row of the stack table at the bottom of this body —
the same commit is **533 files, +22,715, −8,344**. The difference is five rename pairs that GitHub
collapses into five single entries; both numbers are correct and they are reconciled at the end of
"How GitHub will render this".

**Read the rendering note below before you open the Files-changed tab.** This is the only PR in this
stack — other than the planning record at the top — that exceeds *both* of GitHub's diff-rendering
budgets. You will not be shown all of it, and the section "How GitHub will render this" tells you what
to do about that.

---

   zero files.

not asserted: every off-diagonal cell of the pairwise overlap matrix is `0`, and re-verified after the
two slices above this one were cut — the nine cut slices plus the untouched remainder produce tree
`276d89a94`, which is the development branch's tree hash, and the partition arithmetic closes at a gap
of `0`: 252 + 97 + 119 + 162 + 195 + 533 + 214 + 330 + 2,394 = **4,296**, the comparable file total.
The remainder also decomposes independently, without reference to that catch-all measurement:
39 + 37 + 2,318 = 2,394.

## What you are looking at

`apps/frontend/src/lib` — the frontend's whole library layer, and the largest single reviewing surface
in the stack. **212 files added, 60 deleted, 261 modified.**

| --- | ---: | --- |
| `i18n/` | 154 | 102 added, 52 modified. The message catalogues and the `t()` wrapper. |
| `contexts/` | 108 | The Svelte 5 rune-native context layer that replaces the old `svelte/store` seam — app, voter, candidate, admin, data, layout, i18n, component, auth. |
| `components/` | 102 | The base component library (Button, Input, Select, Modal, Expander, Tabs, Video, …). |
| `api/` | 67 | The Supabase adapter (24 files), the universal adapter base, and the data/writer seams. |
| `dynamic-components/` | 48 | The data-aware components — entity cards and lists, filters, feedback, question inputs. |
| `utils/`, `candidate/`, `server/`, `admin/`, `auth/`, `types/`, `supabase/`, `_guards/` | 54 | Route building, sanitisation, logging, the admin job pipeline, and the lint-guard fixture. |

By extension: **278 `.ts`**, **110 `.svelte`**, **137 `.json`**, **8 `.md`**.

Three things are worth going to directly:

- **`contexts/` is where the milestone's central change lives.** The old `svelte/store` bridges are
  gone; contexts are rune-native classes. Two of their properties behave in a way no compiler, lint
  rule or type can enforce, and the project has a documented incident where a violation shipped — see
  the next section, which is the single most useful thing in this PR body.
- **`api/adapters/supabase/` (24 files) is the entire data path.** It replaces the deleted Strapi
  provider with a Supabase adapter built on a mixin that threads SvelteKit's load `fetch` through so
  the same code runs on the server and in the browser against one session.
- **`i18n/wrapper.ts` decides what a user sees when a translation is missing.** `t()` checks runtime
  overrides, then a compiled Paraglide message, and **on a throw or a miss returns the key itself** —
  so an i18n defect surfaces to a user as a raw dotted identifier rather than as an error. That
  behaviour is why PR 9 (the message catalogues) is worth reading as a correctness surface and not as
  a translation dump.

## The two reactivity rules a reviewer cannot see, and how this PR was checked against them

The contexts expose two classes of property with **different reactivity semantics under
destructuring**, and getting it wrong produces a component that renders once and then silently stops
updating. Nothing enforces either rule — not `tsc`, not ESLint, not a test.

1. **Value-replacing reactive accessors** (`appSettings`, `locale`, and the array getters like
   `selectedElections`, `opinionQuestions`, `matches`). Destructuring one invokes the getter **once**
   at component init and binds the value; later reads are reads of a dead local. They must be read as
   `ctx.X`, optionally aliased through `$derived(ctx.X)`.
2. **Identity-stable accessors** — `dataRoot`, whose object reference *never changes* and whose only
   reactive signal is a private version counter bumped on update. For this one a `$derived` alias is
   **also** wrong: the alias recomputes on every bump, yields the same reference every time, and
   Svelte 5's referential-equality rule then skips downstream notification. The consumer keeps the
   pre-mount snapshot. Warm navigation masks it; a cold direct-URL entry exposes it. The only safe
   form is `ctx.dataRoot.<prop>` read **directly inside the consuming tracking scope**.

Rather than sample, every consumer in this slice was enumerated mechanically and read individually:

| Check | Result |
| --- | ---: |
| context destructuring sites in `src/lib` | **60** |
| of those, destructuring a reactive accessor | **0** |
| `dataRoot` consumers in `src/lib` | **5** |
| of those, bound to an intermediate `$derived` alias | **0** |

All five `dataRoot` consumers read through the accessor inside their own tracking scope and say so in
a comment; one defers the read into the consumer with an explicit thunk (`dataRoot: () => DataRoot`,
called inside `$derived.by`). **This slice is clean, and clean deliberately.** The same scan was then
extended past the pathspec boundary into the routing surface, because the same contexts are consumed
there — and that is where three findings were: one genuine defect and two look-alikes. All three are
handled in PR 8, which is where those files live.

recorded verdict: **12 general items plus the three Supabase Adapter items**, all terminal, no blanks.
The adapter block applies to **this slice and to no other in the stack**, so a miss here has no second
chance and all three are proven by enumeration rather than by sample:

| Adapter item | Checked | Conforming |
| --- | ---: | ---: |
| Adapter classes use the mixin's `fetch`-carrying initialiser | 4 concrete classes | **4** |
| Row mapping goes through the shared column map | 11 read-mapping sites; 24 files scanned for hand-rolled case conversion | **11**, and **0** hand-rolled |
| Route guards use the safe session accessor | 9 safe call sites | **9**, and **unsafe-in-a-guard = 0** |

That last number is the one that matters, because `getSession()` compiles exactly as readily as
`safeGetSession()` and nothing anywhere enforces the difference. Both bare `getSession()` occurrences
are named individually rather than netted: one **is** `safeGetSession`'s own implementation and
validates via `getUser()` on the next line, and the other is a data-read method whose `role` does
reach an authorization decision — but behind a `safeGetSession` gate that runs first and redirects.

Where an item is enumerable it is reported as a count with the command behind it, not as an
impression:

| Check | Checked | Result |
| --- | ---: | ---: |
| `{@html}` occurrences / of those wrapped in the sanitiser | 533 files | **19 / 19** — difference **0** |
| `eval(` / `new Function(` | 533 files | **0** |
| secret-shaped literals (`eyJ…`, `password`/`secret`/`api_key`/`service_role` assignments) | 533 files | **0** |
| `any` in type position | 533 files | **6**, in 4 files, **all 6 now documented** |
| `@ts-ignore` | 533 files | **0** (one `@ts-expect-error`, with a four-line note) |
| `svelte/store` imports | 533 files | **0** (the frontend's own lint config makes one an error) |
| `svelte-ignore` directives / of those suppressing an a11y rule / carrying a rationale | 533 files | **26 / 9 / 9** |
| tracking-event call sites whose event name is a union member | 13 | **13** |
| visual baselines regenerated | this slice's commits | **0** |

**Nine defects were fixed on the development branch *before* this slice was cut**, so what you are

- **Four component usage examples documented an API that no longer exists.** Their `@component`
  docblocks showed `bind:selected={$selectedConstituencies}` and `answer={$voterAnswers[question.id]}`
  — the `$store` auto-subscription syntax for the seam this milestone deletes, which the frontend's
  own ESLint config now makes an **error**. A reader copying the documented usage would write code
  that cannot compile. One was wrong a second, independent way: the prop is typed `Id | null` and the
  example passed an `Answer`.
- **Five `console.*` statements bypassed the admin job controller, two at ERROR level.** One shipped
  an entire `Election` object and another every candidate-nomination id, unguarded, on every run of a
  job an admin triggers from the UI — neither an error, and neither reaching the admin. The two were
  deleted and the three progress dumps moved onto the controller, which writes where an admin can see
  them. `console.*` under `server/admin` is now 0.
- **A tracked file whose *name* was editor debris** — `README.md 21-40-30-014.md`, a space and a
  timestamp-shaped suffix. No content grep could ever have found it. Renamed; a filename scan over
  every file under this pathspec finds no second instance.
- **Three `Array<any>` mixin rest parameters were undocumented, and the reason they were invisible is
  the finding.** They do not trip the lint gate because `no-explicit-any` is configured with
  `ignoreRestArgs: true` — they were exempted by configuration, not judged conforming. The comments
  now say so.

## Three things a reviewer should know are open

None is deferred quietly; each is recorded with its reason.

- **The accessibility gate cannot see most of this slice.** The axe scan reaches **5 distinct URLs ×
  2 themes** — measured, not estimated — and it reaches components only *transitively*, through what
  those five URLs happen to render. Of this slice's 110 components, any not mounted on `/`,
  `/elections`, `/constituencies`, `/questions` or `/results` is **not examined by that gate at all**,
  including every candidate-app and admin-app component. The raw-translation-key scan has exactly the
  same reach, because its only call site is inside the axe helper. And axe is a static-DOM auditor: it
  does not tab through the UI, so the **keyboard** half of the requirement is uncovered even on the
  five URLs it does reach — and this slice contains the hand-rolled interactive primitives that make
  that requirement real (`Select` with `aria-activedescendant` and index navigation, the radio-group
  focus-out handling in `QuestionChoices`, `Tabs`, `Expander`, `Modal`). Widening the route table and
  adding a keyboard gate is net-new test authorship and is scoped to a later phase.
- **A real duplication is shipping unfixed.** `EntityListControls` and `EntityListWithControls` carry
  the same three filter handlers — including an identical tracking guard and an identical five-line
  active-filter chain — followed by near-identical control markup, and the extraction target already
  exists beside them. Not fixed because extracting a shared handler is code restructuring, which this
  review pass deliberately excludes, and because it is behaviour-adjacent in the filter path with no
  end-to-end gate available to prove it safe. The remedy is recorded so nobody has to re-derive it.
- **The write half of the row-mapping contract has no production consumer.** `mapRowToDb` and the
  reverse property map are exported and unit-tested with **zero call sites** anywhere in the
  repository, and the write that would want them cannot use them: the shared column map lacks the
  operational columns of the admin-jobs table, so the writer that would need `mapRowToDb` correctly
  spells twelve snake_case keys by hand. Both docblocks now state that and the precondition for future
  use. Widening the map is a change to the package in PR 4, which is already open.

## Why no checks run here

`ship/v0.2-akita-05-e2e-tests`, a sibling branch, so **no checks run — and that is correct behaviour,
not a misconfiguration.** Only PR 1 of the stack fires CI at all. Asserted, not predicted:

**Do not read that as "the jobs exist and were skipped."** The workflow file at this branch's head is

**`Setup Yarn 4.6`** with

```
```

**What was actually run, locally, against the development branch tip that these slices are cut from:**
`yarn build` 14/14 packages, `yarn test:unit` **1,522 tests across 149 files**, `yarn lint:check`
**0 errors** — each measured under a forced (uncached) run. **The 43 end-to-end specs were not run**;
that suite needs a dev server and a seeded local database, and this project's own rule is that a
did-not-run E2E test counts as a failure rather than a pass, so **no green suite is claimed for this
PR**. The suite is run once, in full, against the post-review tip, and that run — not per-PR CI — is
the stack's collective proof.

**This PR exceeds both of GitHub's rendering budgets, and it is the only one in the stack that does
apart from the planning record at the top.**

| Budget | This PR, as GitHub counts it | Over by |
| --- | --- | --- |
| ~300 files before the Files-changed tab truncates | **528 files** | 228 |
| 20,000 changed lines before diffs collapse behind "Load diff" | **30,729** (22,550 + 8,179) | 10,729 |

**Reconciling the two file counts, because you will see one number here and another in the stack table
below.** GitHub applies rename detection and reports `528 files, +22,550, −8,179`; the stack table's
rows are measured with rename detection *off*, which gives `533 files, +22,715, −8,344` for this same
commit. Five files in this slice moved path with their content close enough to match, so rename
detection renders each as one entry instead of a delete plus an add — five fewer files, and 165
insertions and 165 deletions that stop being counted twice. Reproduce either with
`git show --shortstat ship/v0.2-akita-06-frontend-lib` (GitHub's view, renames on) or the same command
with `--no-renames` (the table's view). Neither number changes the conclusion: this PR is over both
budgets on either measure.

So: the file list will be cut off, and much of what does render will be collapsed. **That is known and
accepted rather than fixed by re-partitioning**, and the reason is worth stating plainly — `src/lib`
is one subsystem, and every way of splitting it produces a PR whose title needs an *"and also"*. The
alternative to a big honest diff here was several dishonest small ones.

**A guided path, in the order that makes the rest legible:**

1. **`contexts/` first, and `contexts/app/appContext.svelte.ts` first of all.** It is where the rune
   conversion is decided, and it is where the two reactivity rules above are *installed* — the
   version-bridge getter and the spread-safe `defineProperty` accessors. Read the mechanism before
   judging any consumer; the same source line explains both failure modes.
2. **`api/adapters/supabase/` next** (24 files), against the three adapter checks tabulated above.
   `supabaseAdapter.ts` is the mixin; `utils/mapRow.ts` and `utils/toDataObject.ts` are the row path.
3. **`components/` and `dynamic-components/` last, by exception rather than in full.** 212 of them are
   new files and reading all 110 components is not a good use of a reviewer. The nine `a11y_*`
   suppressions, the 19 `{@html}` sites and the four repaired `@component` examples are where the
   interesting decisions are, and all three sets are enumerated above with counts you can re-derive.
4. `i18n/` is 154 files and is mostly catalogue data; `i18n/wrapper.ts` is the one file in it that
   decides behaviour.

`git diff --stat ship/v0.2-akita-05-e2e-tests..ship/v0.2-akita-06-frontend-lib` locally will give you
the whole file list that the web UI will not.

## The stack

`main`.

| **7** | **`ship/v0.2-akita-06-frontend-lib`** ← *you are here* | rewrite the frontend library layer on Svelte 5 runes and the Supabase adapter | **533** | **22715** | **8344** |
| 12 | `ship/v0.2-akita-11-planning` | add the v0.2 planning record and agent configuration | 2318 | 866928 | 100 |

is what a reviewer actually sees — `git show -M --shortstat` reports `1316 files changed, 0
insertions(+), 0 deletions(-)`, and GitHub's own API independently confirms `files=1316, +0, -0` with
1,316 `renamed` entries. Rows 7, 8 and 9 are measured (all three are cut); rows 10–12 are from the
partition's dry run and move slightly as fixes land on the development branch before each slice is
cut.

## Full evidence

**"Slice 06 — `ship/v0.2-akita-06-frontend-lib`"**.

[151-DISPOSITION.md § Slice 06](https://github.com/OpenVAA/voting-advice-application/blob/ship/v0.2-akita-11-planning/.planning/phases/151-ship-v0-2-akita-review-stack/151-DISPOSITION.md#slice-06--shipv02-akita-06-frontend-lib)


apps/frontend/src/lib/api/adapters/apiRoute/apiRouteAdapter.ts
```diff
@@ -10,6 +10,8 @@ import type { ApiGetRoute, ApiPostRoute, ApiRouteReturnType } from './apiRoutes'
  */
 export function apiRouteAdapterMixin<TBase extends Constructor>(base: TBase): Constructor<ApiRouteAdapter> & TBase {
   abstract class WithMixin extends base {
+    // reason: forwards the base constructor's parameters unchanged; typed as the mixin
```
kaljarv, line 13: Remove line breaks from multiline comments everywhere.

apps/frontend/src/lib/api/adapters/supabase/adminWriter/supabaseAdminWriter.ts
```diff
@@ -0,0 +1,98 @@
+import { UniversalAdapter } from '$lib/api/base/universalAdapter';
+import { supabaseAdapterMixin } from '../supabaseAdapter';
+import type { DataApiActionResult } from '$lib/api/base/actionResult.type';
+import type { InsertJobResultOptions, SetQuestionOptions } from '$lib/api/base/dataWriter.type';
+
+/**
+ * Supabase implementation of admin-specific write operations.
+ *
+ * This class extracts admin methods that were erroneously placed in DataWriter
```
kaljarv, line 9: Remove historical narrative and line breaks.

apps/frontend/src/lib/api/adapters/supabase/adminWriter/supabaseAdminWriter.ts
```diff
@@ -0,0 +1,98 @@
...
+    templates,
+    recipientUserIds,
+    from,
+    dryRun
+  }: {
+    templates: Record<string, { subject: string; text: string; html: string }>;
+    recipientUserIds: Array<string>;
+    from?: string;
+    dryRun?: boolean;
+  }): Promise<{ type: 'success'; sent: number; failed: number; results: Array<unknown> }> {
```
kaljarv, line 80: Add typing for return results if possible.

apps/frontend/src/lib/api/adapters/supabase/dataProvider/supabaseDataProvider.ts
```diff
@@ -0,0 +1,616 @@
...
+    if (error) {
+      if (error.code === 'PGRST116') return {}; // No rows -- return empty settings
+      throw new Error(`getAppSettings: ${error.message}`);
+    }
+
+    const settings = (data?.settings ?? {}) as Record<string, unknown>;
+    const locale = options?.locale ?? this.locale;
+
+    // Localize notification title and content fields
+    if (settings.notifications && typeof settings.notifications === 'object') {
```
kaljarv, line 60: Add validation for the settings object (and all other typed jsonbs) so we don't need typecasts.

apps/frontend/src/lib/api/adapters/supabase/dataProvider/supabaseDataProvider.ts
```diff
@@ -0,0 +1,616 @@
...
+   */
+  protected async _getAppCustomization(options?: GetAppCustomizationOptions): Promise<DPDataType['appCustomization']> {
+    const { data, error } = await this.supabase.from('app_settings').select('customization').limit(1).single();
+
+    if (error) {
+      if (error.code === 'PGRST116') return {} as AppCustomization;
+      throw new Error(`getAppCustomization: ${error.message}`);
+    }
+
+    const raw = (data?.customization ?? {}) as Record<string, unknown>;
```
kaljarv, line 92: Ditto

apps/frontend/src/lib/api/adapters/supabase/dataProvider/supabaseDataProvider.ts
```diff
@@ -0,0 +1,616 @@
...
+          color: row.entity_color,
+          image: row.entity_image,
+          sort_order: row.entity_sort_order,
+          subtype: row.entity_subtype,
+          custom_data: row.entity_custom_data
+        };
+        const entityObj = toDataObject(entityRow, locale, this.defaultLocale);
+        const entityType = row.entity_type;
+
+        // Explicitly-typed shared DataObject fields (localized name/short_name/
```
kaljarv, line 361: Aren't the fields already typed by toDataObject? They should be. Also, remove line breaks from comment.

apps/frontend/src/lib/api/adapters/supabase/dataProvider/supabaseDataProvider.ts
```diff
@@ -0,0 +1,616 @@
...
+
+    // 1. Fetch categories
+    const { data: catData, error: catError } = await this.supabase
+      .from('question_categories')
+      .select('*')
+      .order('sort_order');
+    if (catError) throw new Error(`getQuestionData (categories): ${catError.message}`);
+
+    let categories = (catData ?? []).map((row) => {
+      const obj = toDataObject(row as Record<string, unknown>, locale, this.defaultLocale);
```
kaljarv, line 511: Make sure we need these smelly typecasts nowhere.

apps/frontend/src/lib/api/adapters/supabase/dataProvider/supabaseDataProvider.ts
```diff
@@ -0,0 +1,616 @@
...
+
+    // The `get_nominations` RPC accepts a single uuid per election/constituency.
+    // When the caller passes arrays (multi-election voter flow — see
+    // `(voters)/(located)/+layout.ts` which threads URL `electionId` /
+    // `constituencyId` arrays through verbatim), fan out into one RPC per
+    // (election, constituency) pair and concatenate the results. Picking
+    // `[0]` only — the prior shape — silently dropped the other elections'
+    // nominations and broke the multi-election partial-coverage dialog gate
+    // (variant-constituency.spec.ts:237).
+    const electionIds: Array<string | null> = options?.electionId
```
kaljarv, line 245: Extract to a convertFilterValue or similar helper.
Extend options to include electionRound filtering and edit the rpc.

apps/frontend/src/lib/api/adapters/supabase/dataProvider/supabaseDataProvider.ts
```diff
@@ -0,0 +1,616 @@
...
+
+    return results;
+  }
+
+  /**
+   * Fetch question categories and questions. Localizes choice labels for
+   * choice-type questions, maps category_type to type on categories,
+   * and filters categories by electionId when specified.
+   */
+  protected async _getQuestionData(options?: GetQuestionsOptions): Promise<DPDataType['questions']> {
```
kaljarv, line 499: Enable filtering by election, constituency and election round.
Create a get_questions rpc to that effect, which returns both categories and their questions.

apps/frontend/src/lib/api/adapters/supabase/dataWriter/supabaseDataWriter.ts
```diff
@@ -0,0 +1,424 @@
...
+  protected async _resetPassword({ password }: { password: string; code: string }) {
+    // Called after recovery session is established via auth callback.
+    // The `code` param is unused; Supabase uses the recovery session.
+    const { error } = await this.supabase.auth.updateUser({ password });
+    if (error) throw new Error(error.message);
+    return { type: 'success' as const };
+  }
+
+  protected async _setPassword({ password }: { password: string; currentPassword: string; authToken: string }) {
+    // currentPassword and authToken are WithAuth compatibility shims -- ignored by Supabase.
```
kaljarv, line 84: Edit the interface shape so that the withAuth shim is no longer needed if it has no use with supabase.

apps/frontend/src/lib/api/adapters/supabase/dataWriter/supabaseDataWriter.ts
```diff
@@ -0,0 +1,424 @@
...
+      termsOfUseAccepted: data.terms_of_use_accepted ?? null,
+      // reason: JSONB columns return Json (structural superset of StoredImage); parseStoredImage runtime-guards on .path
+      image: parseStoredImage(data.image as Json as unknown as StoredImage | null, constants.PUBLIC_SUPABASE_URL)
+    } as unknown as LocalizedCandidateData;
+  }
+
+  ////////////////////////////////////////////////////////////////////
+  // ADMIN METHODS
+  // TODO: Primary access point is SupabaseAdminWriter. These
+  // implementations satisfy the abstract contract on UniversalDataWriter.
```
kaljarv, line 381: Remove these from here and the abstract interface.

apps/frontend/src/lib/api/adapters/supabase/utils/getLocalized.ts
kaljarv, line 1: These and the test should be colocated with packages/app-shared/src/data/localized.type.ts and use the types there.

apps/frontend/src/lib/api/utils/auth/providers/authConfig.ts
kaljarv, line 1: The should be either included in the provider implementations, or split into provider-specific files so they're not interdependent.

apps/frontend/src/lib/api/utils/auth/providers/index.ts
```diff
@@ -0,0 +1,42 @@
...
+ * Get the active identity provider based on environment configuration.
+ *
+ * Reads `PUBLIC_IDENTITY_PROVIDER_TYPE` from public constants and returns the
+ * corresponding provider implementation. Defaults to `'signicat'` when the
+ * env var is unset or empty.
+ *
+ * @throws {Error} If `PUBLIC_IDENTITY_PROVIDER_TYPE` contains an unsupported value.
+ */
+export function getActiveProvider(): IdentityProvider {
+  const providerType = (constants.PUBLIC_IDENTITY_PROVIDER_TYPE || 'signicat') as ProviderType;
```
kaljarv, line 31: Remove the default with no historical mentions.

apps/frontend/src/lib/api/dataProvider.ts
```diff
@@ -1,19 +1 @@
...
-
-let module: Promise<{ dataProvider: UniversalDataProvider }>;
-
-const { type } = staticSettings.dataAdapter;
-
-switch (type) {
-  case 'strapi':
-    module = import('./adapters/strapi/dataProvider');
-    break;
-  case 'local':
```
kaljarv, line 12: We should add reintroducing the local adapter as a follow-up task.

apps/frontend/src/lib/api/README.md
```diff
@@ -1,12 +1,12 @@
...
-> See also the online doc [Data API](https://openvaa.org/developers-guide/frontend/data-api) (or [locally](/docs/src/routes/developers-guide/frontend/data-api/+page.md))
+> See also the online doc [Data API](https://openvaa.org/developers-guide/frontend/data-api) (or [locally](</apps/docs/src/routes/(content)/developers-guide/frontend/data-api/+page.md>))
 
 This directory contains the Data API implementation for the OpenVAA frontend, including:
 
 - Data providers for reading public data
 - Feedback writers for submitting feedback
 - Data writers for authenticated data operations
-- Multiple implementations (Strapi, API routes, local server)
+- Supabase adapter for all data access (local adapter available for static data)
```
kaljarv, line 10: Not accurate now with local disabled.

apps/frontend/src/lib/auth/getUserData.ts
```diff
@@ -1,54 +1,33 @@
...
+    if (!parentData.session) return undefined;
+  }
 
   const dataWriter = await dataWriterPromise;
   dataWriter.init({ fetch });
 
-  const userData = await dataWriter.getBasicUserData({ authToken }).catch((e) => {
+  // authToken is ignored by Supabase adapter -- session is cookie-based
+  const userData = await dataWriter.getBasicUserData({ authToken: '' }).catch((e) => {
     logDebugError(`Error fetching user data: ${e?.message ?? 'No error message'}`);
```
kaljarv, line 31: Add a blocking follow-up for renaming and refactoring logDebugError to a smarter method with structured, pino and OTL ready conformant output.

apps/frontend/src/lib/candidate/components/passwordSetter/PasswordSetter.svelte
```diff
@@ -47,30 +49,41 @@ Contains the dynamic `PasswordValidator` component.
...
 
-  $: valid = !!(password && passwordConfirmation && validPassword && password === passwordConfirmation);
-  $: if (!validPassword) {
-    errorMessage = $t('candidateApp.setPassword.passwordNotValid');
-  } else if (password !== passwordConfirmation) {
-    errorMessage = $t('candidateApp.setPassword.passwordsDontMatch');
-  } else {
-    errorMessage = undefined;
-  }
+  $effect(() => {
```
kaljarv, line 55: Check if these could be $deriveds.

If they can, check all other $effect calls for similar patterns.

apps/frontend/src/lib/components/alert/Alert.svelte
```diff
@@ -83,49 +79,49 @@ Show a non-model alert or dialog that appears at the bottom of the screen.
...
-      <slot name="actions" />
+    {#if actions}
+      {@render actions()}
     {:else}
-      <Button on:click={closeAlert} color="warning" text={$t('common.close')} class="-mt-[1rem] sm:mt-0" />
+      <Button onclick={closeAlert} color="warning" text={t('common.close')} class="-mt-[1rem] sm:mt-0" />
     {/if}
   </div>
-  <button on:click={closeAlert} class="btn btn-circle btn-ghost btn-sm absolute right-2 top-2">
+  <button onclick={closeAlert} class="btn btn-circle btn-ghost btn-sm absolute top-2 right-2">
```
kaljarv, line 117: Use closest available semantic class, e.g. top-sm.

apps/frontend/src/lib/components/input/Input.svelte
```diff
@@ -282,6 +303,23 @@ The input itself is wrapped in multiple container elements, the outermost of whi
...
+      // Only update the value if it's an empty string or a valid email
+      const currentValue = currentTarget.value.trim();
+      if (currentValue === '') {
+        value = '';
+      } else {
+        if (!EMAIL_REGEX.test(currentValue)) return handleError('components.input.error.invalidEmail');
+        value = currentValue;
+      }
+
+      // Number — coerce the DOM string value to a real JS number (or undefined when cleared).
```
kaljarv, line 316: Remove comment, the frontend components should not reference data provider implementations.

apps/frontend/src/lib/components/input/MultipleTextInput.svelte
kaljarv, line 1: This should be either incorporated into Input like the other types or only the input part extracted and imported into Input and the same extraction done to other more complicated types in Input.

The multilingual option should also be implemented at the same time such that each text item behaves much the same as a normal multilingual text item.

apps/frontend/src/lib/components/questions/QuestionChoices.svelte
kaljarv, line 1: Add as a follow-up blocking task for me to UAT:
- BooleanInput
- multi-select choices

apps/frontend/src/lib/contexts/app/appContext.svelte.ts
```diff
@@ -0,0 +1,477 @@
...
+    // producers/contexts.
+    ////////////////////////////////////////////////////////////////////
+
+    // Forward dataCtx.dataRoot as a BARE own-enumerable reactive accessor (see phase 113
+    // ): read the now-bare `dataContext.dataRoot` accessor each access so the
+    // `#version` reactive re-read inside it stays live, and install via
+    // Object.defineProperty(enumerable:true) so it survives the downstream
+    // `{ ...appContext }` spreads. (Cannot go through the plain `Object.assign` value
+    // copy below — that would snapshot the DataRoot once, losing reactivity.)
+    Object.defineProperty(this, 'dataRoot', {
```
kaljarv, line 335: This ad hoc rollup is fixed in the last branch of the stack.

apps/frontend/src/lib/contexts/app/tracking/trackingService.svelte.ts
```diff
@@ -0,0 +1,221 @@
...
+import type { TrackingEvent } from './trackingEvent.type';
+import type { TrackingHandler, TrackingService } from './trackingService.type';
+
+/**
+ * The pure-rune internal shape of the tracking service. The `appContext` seam
+ * owns the store conversion of the store-shaped properties (`sendTrackingEvent`,
+ * `sessionId`, `shouldTrack`) declared on the exported `TrackingService` type —
+ * this producer exposes them as rune handles.
+ */
+export type RuneTrackingService = Omit<TrackingService, 'sendTrackingEvent' | 'sessionId' | 'shouldTrack'> & {
```
kaljarv, line 17: Investigate whether we can have just one tracking service type and implementation and not have these two layers. Any properties only used by the tracking service internally need not necessarily be easily accessible. Therefore we can consider making the consumer facing interfaces' definition narrower. The sessionId, for example, need not be used by the consumers, just the service internally.

apps/frontend/src/lib/contexts/app/reactiveHandle.type.ts
kaljarv, line 1: The more logical place for these could be in contexts/utils

apps/frontend/src/lib/contexts/candidate/candidateContext.svelte.ts
```diff
@@ -0,0 +1,615 @@
...
+        this.#selectedConstituencies = removeDuplicates(
+          current.nominations.nominations.map((n) => dr.getConstituency(n.constituencyId))
+        );
+      } catch (e) {
+        logDebugError(`[candidateContext selectedConstituencies] Error fetching constituency: ${e}`);
+        this.#selectedConstituencies = [];
+      }
+    });
+
+    $effect(() => {
```
kaljarv, line 355: Is it possible to use a common utility for this and the same block in VoterContext?

apps/frontend/src/lib/contexts/utils/SettingsOverlay.svelte.ts
kaljarv, line 1: Rename this and the test file to settingsOverlay

apps/frontend/src/lib/contexts/voter/voterContext.svelte.ts
```diff
@@ -0,0 +1,686 @@
...
+// which produces fresh query-param arrays even when content is unchanged
+// (e.g., drawer open/close adds /candidate/[id] route segments while
+// electionId search param is identical). Without this guard, every
+// navigation cascaded selectedElections → nominationAndQuestionState →
+// filterState, rebuilding FilterGroup instances and dropping any active
+// filter rules — surfaced during manual smoke (see phase 64) as "filter badge
+// disappears after closing candidate drawer". Svelte 4 stores absorbed this
+// via `writable.set()`'s no-op-write skip; Svelte 5 raw `$state` writes
+// need an explicit equality check.
+function sameRefs<TItem>(a: ReadonlyArray<TItem>, b: ReadonlyArray<TItem>): boolean {
```
kaljarv, line 37: Move to the bottom of the file or in utils.

apps/frontend/src/lib/dynamic-components/entityCard/EntityCard.svelte
```diff
@@ -53,130 +53,186 @@ This is a dynamic component, because it accesses the `dataRoot` and other proper
...
+        route: 'ResultEntity',
+        entityTab: type === 'candidate' ? 'candidates' : type === 'alliance' ? 'alliances' : 'organizations',
+        entity: type,
+        id,
+        nominationId: unwrapped.nomination?.id
+      });
 
     // The questions and possible submatches to display in the card
     // TODO: Add support for all entity types by expanding the setting type to cover all of them
+    let qs: CardQuestions | undefined;
```
kaljarv, line 126: Don't use variable names as terse as this (and scs etc.).

The only place to use such is in small ad hoc closures like .filter((o) => o.foo)).

apps/frontend/src/lib/dynamic-components/entityCard/EntityCardAction.svelte
```diff
@@ -11,12 +9,12 @@ TODO[Svelte 5]: Maybe convert into `$snippet`.
 
 ### Slots
 
-– default: The contents to wrap.
+\u2013 default: The contents to wrap.
```
kaljarv, line 12: Don't encode dashes in comments, just use a hyphen

apps/frontend/src/lib/dynamic-components/entityCard/EntityCard.svelte
```diff
@@ -53,130 +53,186 @@ This is a dynamic component, because it accesses the `dataRoot` and other proper
...
+    if (variant !== 'subcard') {
+      c += ' rounded-md bg-base-100 px-md py-16';
+      if (parsed.action) c += ' text-neutral transition-shadow ease-in-out hover:shadow-xl';
+    }
+    return c;
+  });
 </script>
 
 <!-- If there are no subcards, we make the whole card clickable... -->
 <EntityCardAction
```
kaljarv, line 220: EntityCardAction is a pre-snipper era workaround, so see if we can just use a snippet here and remove the separate component.

apps/frontend/src/lib/dynamic-components/entityList/EntityListWithControls.helpers.ts
kaljarv, line 1: Rename to just helpers.ts

apps/frontend/src/lib/i18n/init.ts
```diff
@@ -46,155 +40,42 @@ if (!defaultLocale) {
...
-
-/**
- * Return the correct string for the `locale` using soft-matching from the supplied `LocalizedString` object.
- *
- * @param strings - An object with locale-translation key-value pairs or a raw string
- * @param locale - The target locale
- * @returns The translalated string or ''
+ * Return the correct string for the locale using soft-matching from the supplied LocalizedString object.
  */
 export function translate(strings: LocalizedString | string | undefined | null, locale?: string | null): string {
```
kaljarv, line 52: Check whether these utils are any longer needed after the extraction of translation utils to app-shared in an earlier review comment.

apps/frontend/src/lib/utils/constants.ts
```diff
@@ -7,6 +7,9 @@ export const constants = {
   PUBLIC_SERVER_FRONTEND_URL: env.PUBLIC_SERVER_FRONTEND_URL ?? '',
   PUBLIC_IDENTITY_PROVIDER_CLIENT_ID: env.PUBLIC_IDENTITY_PROVIDER_CLIENT_ID ?? '',
   PUBLIC_IDENTITY_PROVIDER_AUTHORIZATION_ENDPOINT: env.PUBLIC_IDENTITY_PROVIDER_AUTHORIZATION_ENDPOINT ?? '',
+  PUBLIC_IDENTITY_PROVIDER_TYPE: env.PUBLIC_IDENTITY_PROVIDER_TYPE ?? 'signicat',
```
kaljarv, line 10: Remove default.

apps/frontend/src/lib/utils/getAllianceSummary.ts
kaljarv, line 1: Rename to alliances.ts

apps/frontend/src/lib/utils/multiChoiceValidity.ts
```diff
@@ -0,0 +1,31 @@
+/**
+ * The single source of truth for the multi-choice categorical
+ * selection-count validity used by `OpinionQuestionInput` (and, transitively,
+ * the voter persistence gate + candidate Save gate).
+ *
+ * A selection is valid (a saveable answer) when its `count` falls inclusively
+ * within `[effectiveMin, effectiveMax]`, where:
+ * - `effectiveMin = minSelections ?? 1` — zero selections is ALWAYS
```
kaljarv, line 8: If so, minSelection should be checked to be > 0.

kaljarv (commented):
Checked.

PR #870, opened 2026-08-17
8/12 feat: rewrite the frontend app shell and the voter and candidate routing surface

kaljarv:
# 8/12 · `feat: rewrite the frontend app shell and the voter and candidate routing surface`

**Stack position: 8 of 12.** Base: `ship/v0.2-akita-06-frontend-lib`. Head: `ship/v0.2-akita-07-frontend-routes`.
**GitHub will show you 165 files changed, 7,593 insertions, 5,542 deletions.** Measured without
rename detection — the convention used for every row of the stack table at the bottom of this body —
the same commit is **214 files, +10,319, −8,268**. The gap is large here and it is good news for a
reviewer: GitHub recognises 49 of this slice's moved route files as renames rather than as
delete-plus-add pairs. Both numbers are correct and they are reconciled at the end of "How GitHub will
render this".

**This PR is the request path.** It holds every route in the application, the server hooks that wrap
them, the API endpoints they call, and the app shell that frames them. It is also where the locale
segment leaves the URL: 97 of its 99 deletions are the old `[[lang=locale]]` route tree, re-added
without that segment.

---

   zero files.

not asserted: every off-diagonal cell of the pairwise overlap matrix is `0`, and the nine cut slices
plus the untouched remainder produce tree `276d89a94`, which is the development branch's tree hash.
The partition arithmetic closes at a gap of `0`:
252 + 97 + 119 + 162 + 195 + 533 + 214 + 330 + 2,394 = **4,296**, the comparable file total. The
remainder also decomposes independently, without reference to that catch-all measurement:
39 + 37 + 2,318 = 2,394.

## What you are looking at

`apps/frontend/src/routes`, `apps/frontend/src/params`, and the seven `apps/frontend/src/` shell files.
**109 files added, 99 deleted, 6 modified.**

| --- | ---: | --- |
| `routes/[[lang=locale]]/**` | 97 | **All deletions.** The old locale-parameterised route tree. |
| `routes/candidate/**` | 33 | The candidate app: login, registration, password reset, the five preregistration steps, and the `(protected)` profile/questions/preview/settings set. |
| `routes/(voters)/**` | 25 | The voter app, with the election- and constituency-dependent pages under `(voters)/(located)/`. |
| `routes/admin/**` | 17 | **New in v0.2.** The admin app — login plus four `(protected)` pages driving the LLM job features. |
| `routes/api/**` | 15 | **New in v0.2 as a tree.** 17 endpoints at this head: the six admin job endpoints, the five OIDC endpoints of the bank-authentication exchange, plus auth, preregistration, data, feedback and cache. |
| `routes/*` (app shell) | 15 | `Layout.svelte`, `Header.svelte`, `MainContent.svelte`, `SingleCardContent.svelte`, `Banner.svelte`, `MaintenancePage.svelte`, `+error.svelte`, the root `+layout`, their type files, `README.md`, and `loginRedirectTarget.ts`. |
| `params/` | 6 | Two matchers added with tests (`etPl`, `etSg`); two deleted (`locale`, `entityType`). |
| `src/` shell | 7 | `hooks.server.ts`, `hooks.ts`, `app.html`, `app.d.ts`, `app.css`, `error.html`, `tailwind-theme.css`. |

**`hooks.server.ts` is deliberately in this PR rather than in the configuration PR.** It is the
SvelteKit server hook that creates the per-request Supabase client, installs the validated session
accessor on `event.locals`, resolves the locale, and runs the candidate authentication redirect gate —
the most authorization-relevant file in the frontend. It was moved here during partitioning
specifically so it would be reviewed as part of the request path, alongside the routes it wraps,
instead of beside `turbo.json` and `yarn.lock`. The other six `src/` shell files came with it.

Three things are worth going to directly:

- **`hooks.server.ts`** — three sequenced handlers, in a load-bearing order (`supabaseHandle` first so
  everything downstream can use `event.locals.supabase`). Its `safeGetSession` is the accessor the
  nine route guards in this tree and in PR 7 call; it reads the session and then **validates it with a
  separate user lookup**, returning nulls on error. `hooks.ts` is the reroute hook that strips the
  locale prefix, which is what allows the route tree to carry no locale segment.
- **`routes/api/admin/jobs/**` (6 endpoints)** — each one starts with the same role check and returns
  403 before doing anything else. Enumerated below.
- **`(voters)/(located)/results/[[electionTab]]/+layout.svelte`** — the most intricate component in the
  tree: URL-as-single-source-of-truth, a four-segment optional-parameter route, drawer-first paint,
  and the file where this PR's one genuine reactivity defect was found and fixed.

Backend, Supabase Adapter, Edge Functions) do not apply — this slice touches none of their paths.

**One property of this slice is worth stating because no other slice in the stack has it: the diff is
the entire surface.** Elsewhere in this stack a file can be byte-identical across the layout move,
which means PR 1 renders it as a rename and no later PR's diff contains it — reviewed by nobody. Under
this slice's paths there are **114 tracked files and all 114 are in this diff** (109 added + 6 modified
+ the 99 deletions of their predecessors), and a blob-by-blob comparison against the pre-move tree
finds **0** files unchanged across the move. Nothing here is invisible.

Where an item is enumerable it is reported as a count with the command behind it, not as an
impression:

| Check | Checked | Result |
| --- | ---: | ---: |
| `any` in type position | 214 files | **0** — the cleanest result in the stack on this axis (all 20 `any` matches are the English word in prose) |
| `@ts-expect-error` / `@ts-ignore` | 214 files | **0 / 0** |
| `{@html}` occurrences / of those wrapped in the sanitiser | 214 files | **20 / 20** — difference **0** |
| `eval(` / `new Function(` / `innerHTML` writes | 214 files | **0 / 0 / 0** |
| secret-shaped literals (`eyJ…`, `password`/`secret`/`api_key`/`service_role` assignments) | 214 files | **0** |
| admin job endpoints / of those guarded on the `admin` role before any work | 6 | **6 / 6** |
| server route guards using the validated session accessor / using the raw one | 8 + the hook gate | **9 / 0** |
| visible form controls / of those with a label association or an ARIA name | 214 files | **23 / 23** |
| non-semantic interactive elements (`<div onclick>` and kin) | 36 page routes | **0** |
| `` without `alt` | 214 files | **0** (there is no `` in the route layer at all) |
| positive `tabindex` | 214 files | **1**, the skip link, with a rationale'd suppression |
| `console.*` under `routes/admin/` | after the sweep | **0** |
| visual baselines regenerated | this slice's commits | **0** |

### Accessibility was swept over every route, including the 31 no gate reaches

This is the part of the checklist a reviewer is least able to verify from a diff, so the reach of the
automated gate was measured before anything was claimed. **The end-to-end accessibility scan's route
table resolves to 5 distinct URLs × 2 themes.** This slice contains **36 page routes**. So **31 of
them are examined by no automated gate at all** — every candidate-app route (18), every admin route
(5), and 8 voter routes including `questions/[questionId]`, the page a voter spends most of the
journey on.

The five it does reach: `/`, `/elections`, `/constituencies`, `/questions`, `/results`.

The other 31 were swept by hand for the classes the scan would have applied, and each has a recorded
verdict rather than a shrug. Per-class totals over those 31 routes:

| Class the scan would have caught | Occurrences in the 31 unscanned routes |
| --- | ---: |
| non-semantic element with a click/key handler | **0** |
| positive `tabindex` | **0** |
| `` without `alt` | **0** |
| `autofocus` | **0** |
| `aria-hidden` on a focusable element | **0** |
| empty anchor | **0** |
| `title` used as the only accessible name | **0** |
| heading-level skip | **2 — found, and fixed** |

**The two heading-level skips are the whole argument for doing this by hand.**
`results/[[electionTab]]/statistics/+page.svelte` went straight from the shell's `<h1>` to `<h4>` at
two sites — an axe `heading-order` failure and a WCAG 1.3.1 defect for anyone navigating by heading
level. That route is not in the scan's table, so nothing would ever have reported it. Both are now
`<h2>`, and that this is the correct level was verified rather than assumed: the only candidate for an
intermediate heading is the `<Expander>` wrapping each question, and it renders its title in a plain
`<div>` with no heading role, so the page's real heading tree is the `<h1>` and then these two. The
rendered size is unchanged — the stylesheet sets heading sizes in the base layer, so a size utility on
the element overrides it and leaves the font weight identical.

The structural reason the other 30 come out clean is worth knowing, because it is also a caveat:
**30 of the 31 render through `MainContent` or `SingleCardContent`**, the shell components in this
same PR, which supply the `<h1>` and the `[data-focus-on-nav]` post-navigation focus target. Their
own accessibility structure is therefore load-bearing for the whole app and lives here:
`Layout.svelte` provides the skip link and the `<main>` landmark, `Header.svelte` the menu button with
`aria-expanded` / `aria-controls`, and the root `+layout.svelte` an always-present `aria-live` route
announcer plus an `afterNavigate` focus reset. The 31st route is
`questions/[questionId]/+page.svelte`, a deliberately empty leaf whose rendering is owned by its
parent layout.

**What that sweep does not establish:** it is a static one. Contrast ratios were not measured, and
**keyboard navigation has no automated gate anywhere in this project** — the scan is a static-DOM
auditor and does not tab through the UI, so the keyboard half of the requirement is uncovered on the
five scanned URLs as much as on the other 31. The drawer's close path was traced by hand and does have
proper focus management (open focuses the close button, close returns focus to the opener, and the
overlay that carries a click handler is `aria-hidden` and duplicates a keyboard-reachable control),
but that is one path read carefully, not a gate.

### Six defects were fixed on the development branch before this slice was cut

So what you are reading is the corrected code, and this PR contains no fix of itself.

- **A reactivity defect whose own comment said it was correct.**
  `results/[[electionTab]]/+layout.svelte` destructured two context properties, `appSettings` and
  `dataRoot`, above a comment asserting that both were stable and that destructuring them was fine.
  That was true before those two became bare reactive accessors, and false afterwards. The consequence
  was concrete: the multi-election selector is gated on `{#if dataRoot.elections.length > 1}`, and a
  destructured `dataRoot` takes its version dependency once at component init, so that guard never
  re-evaluated when election data arrived after mount. `appSettings` now reads through a `$derived`
  alias (correct for a value-replacing accessor) and both `dataRoot` reads go directly through the
  context inside their own tracking scope (the only correct form for an identity-stable one, because a
  `$derived` alias would yield the same object reference every time and referential equality then
  suppresses downstream notification). **The stale comment was rewritten too** — a comment asserting
  the wrong rule is how this survived four phases.
- **A second instance of the same defect, in the root layout, which a previous scan had missed.**
  `routes/+layout.svelte` destructured `appSettings` out of the app-context initialiser, whose own
  type declaration reads *"MUST be read off `ctx` (never destructured)"*. The earlier scan matched
  `get*Context()` call sites and the initialiser is not one of those, so a scan that reported one
  violation in this tree had in fact found one of two. Re-run over all **84** context-destructuring
  sites here, both were found and both are fixed.
- **An unvalidated post-login redirect, on both login actions.** Both the candidate and admin login
  form actions read a `redirectTo` field — which arrives from `?redirectTo=` on the login URL and is
  therefore caller-controlled — and interpolated it straight into their redirect target. What made
  that safe was incidental rather than intended: the target is built as `/{locale}/{redirectTo}`, and
  the leading segment stops the value from ever forming a protocol-relative or absolute URL. It did
  not stop a crafted link from steering a just-authenticated user to an arbitrary in-origin path, and
  the voter half of this same PR already validates its equivalent parameter against an allowlist
  *twice*, with a comment naming defence in depth. The guard is now shared by both actions and is
  fail-safe: a rejected value falls through to the app home, which is exactly the existing behaviour
  when no `redirectTo` is present. The application has exactly one legitimate producer of that
  parameter (`hooks.server.ts`), and the predicate was exercised against the four values it actually
  produces plus 20 hostile ones.
- **Nineteen debug `console.*` statements in the two admin form actions** — including three whose only
  purpose was to print `typeof` and `Object.keys` of a value, one block commented `// DEBUG:` that made
  an **extra network round-trip on every admin job start** purely to feed a log line, and one that
  logged the admin's free-text LLM prompt content to the server console at INFO level on every submit.
  This was the route half of a defect already fixed one PR below in the library half of the same two
  features. The two `console.error` calls in the catch blocks now use the project's development-only
  logger, which is what all 22 other route files here use, and both catches already return the message
  to the admin UI.
- **A heading-order skip on an unscanned route** — described above.
- **The routes README omitted two whole route trees this PR adds** — `admin/` and `api/`, 22 of the
  slice's 114 files, i.e. the entire server-side API surface and the entire admin app, missing from the
  one file whose job is to say where routes live. It now describes both, and names the non-route shell
  files, because a reader who knows SvelteKit will otherwise read every `.svelte` in the top directory
  as a route.

Separately, **28 comments the project's automated planning-reference cleanup had left ungrammatical
were repaired.** Nine had collapsed into a fragment with no verb; several had lost the subject a
preposition was governing (`per CONTEXT.`, `per -01 small-fix constraint`); one had had the call
parentheses stripped off a function name one clause before the words "returns an object literal"; and
one had been broken *by the earlier repair pass itself*, while its untouched sibling file still reads
correctly. These were found by diffing the cleanup commits' own before/after line pairs, not by
grepping the tree for damage — the damage signature depends on the original wording, so a grep
under-reports it.

- **The two login form actions are substantially the same action, and the duplication includes the
  security-relevant half.** Normalise the app name and they differ only in a role list, a route
  constant and comment wording — including a verbatim-duplicated hand-rolled decode of the session
  token's payload. Not fixed, because extracting it is code restructuring, which this review pass
  deliberately excludes, and because it is a behaviour change on an authentication path with no
  end-to-end coverage available to prove it safe (`grep -rn redirectTo tests/` matches nothing but an
  unrelated Supabase parameter). The remedy is recorded rather than left to be re-derived. The
  allowlist regex is likewise duplicated at two sites, but there the duplication is *intentional* and
  both copies say so: it is a deliberate second check at the consumption layer.
- **`Expander`'s title is not a heading, and its control has no specific accessible name.** It renders
  the title in a plain `<div>` and its only interactive element is a checkbox whose accessible name is
  a generic "expand or collapse" — so a screen-reader user hears that phrase without learning which
  question it belongs to. That file is in PR 7, which is already open, so fixing it here would mean
  force-pushing a PR under review.

## Why no checks run here

`ship/v0.2-akita-06-frontend-lib`, a sibling branch, so **no checks run — and that is correct

**Do not read that as "the jobs exist and were skipped."** The workflow file at this branch's head is

**`Setup Yarn 4.6`** with

```
```

**What was actually run, locally, against the development branch tip these slices are cut from, after
this PR's six fixes landed:** `yarn build` 14/14 packages, `yarn test:unit` **1,522 tests across 149
files**, `yarn lint:check` **0 errors** — each under a forced, uncached run, and each identical to the
pre-sweep baseline. **The 43 end-to-end specs were not run**; that suite needs a dev server and a
seeded local database, and this project's own rule is that a did-not-run E2E test counts as a failure
rather than a pass, so **no green suite is claimed for this PR**. That matters more here than
elsewhere, because two of this PR's fixes are reactivity fixes and reactivity is what an end-to-end
test is uniquely able to catch. The nearest covering spec is named in the fix's own commit message
along with what it does and does not cover, and the full-suite run against the post-review tip is the
stack's collective proof.

**165 files and 13,135 changed lines as GitHub counts them — comfortably inside both budgets** (the
~300-file limit before the Files-changed tab truncates, and the 20,000-line limit before diffs collapse
behind "Load diff"). The tab will render this PR in full. If you want a reading order:
`hooks.server.ts` and `hooks.ts`, then `routes/api/`, then the app shell files in the top directory,
then the three app trees.

**Reconciling the two file counts, because you will see one number here and another in the stack table
below.** GitHub applies rename detection and reports `165 files, +7,593, −5,542`; the stack table's
rows are measured with rename detection *off*, which gives `214 files, +10,319, −8,268` for this same
commit. The cause is this slice's defining change: 97 of its 99 deletions are the old
`[[lang=locale]]/**` route tree, and each of those files has a counterpart added at the same path minus
the locale segment. **GitHub matches 49 of those pairs as renames**, so it shows one entry per pair
instead of two, and the 2,726 insertions and 2,726 deletions of the matched content stop being counted
twice. Reproduce either with `git show --shortstat ship/v0.2-akita-07-frontend-routes` (GitHub's view,
renames on) or the same command with `--no-renames` (the table's view); the rename-aware status
breakdown is `60 A / 50 D / 6 M / 49 R`.

That is the better rendering for review, not the worse one: a renamed file with a small edit reads as
one diff you can scan, where a delete-plus-add pair asks you to reconstruct the change by eye. For the
48 pairs GitHub does *not* match — the ones whose contents changed too much to pair up — reading the
deleted file beside its added counterpart is still the fastest way to see what actually changed inside
it.

## The stack

`main`.

| **8** | **`ship/v0.2-akita-07-frontend-routes`** ← *you are here* | rewrite the frontend app shell and the voter and candidate routing surface | **214** | **10319** | **8268** |
| 12 | `ship/v0.2-akita-11-planning` | add the v0.2 planning record and agent configuration | 2318 | 866928 | 100 |

is what a reviewer actually sees — GitHub's own API reports `files=1316, +0, -0` with 1,316 `renamed`
entries. Rows 7, 8 and 9 are measured (all three are cut); rows 10–12 are from the partition's dry run
and move slightly as fixes land on the development branch before each slice is cut.

## Full evidence

Every verdict above, with file-and-line citations — including the named list of all 31 unscanned
routes and their individual accessibility verdicts — is in
**"Slice 07 — `ship/v0.2-akita-07-frontend-routes`"**.

[151-DISPOSITION.md § Slice 07](https://github.com/OpenVAA/voting-advice-application/blob/ship/v0.2-akita-11-planning/.planning/phases/151-ship-v0-2-akita-review-stack/151-DISPOSITION.md#slice-07--shipv02-akita-07-frontend-routes)


apps/frontend/src/routes/(voters)/(located)/questions/[questionId]/+page.svelte
Comment (kaljarv) [inline apps/frontend/src/routes/(voters)/(located)/questions/[questionId]/+page.svelte:1]: Check if this is necessary or if some content from the layout can be moved here with the transitions surviving.

apps/frontend/src/routes/(voters)/(located)/questions/+layout.svelte
```diff
@@ -0,0 +1,366 @@
...
+  // Handle `start` query param
+  ////////////////////////////////////////////////////////////////////
+
+  // Assumption A2 / Pitfall 3: kept as `onMount`. The `start` param is only
+  // present on the entry URL of a "start answering from here" deep-link, so the
+  // handler must run once per session — exactly what `onMount` on the persistent
+  // layout provides (the layout mounts once per session, like the leaf did under
+  // SvelteKit's page-reuse). Porting to `afterNavigate`/$effect would re-fire on
+  // every hop and is unnecessary; `onMount` preserves current behavior verbatim.
+  onMount(() => {
```
kaljarv, line 144: Add a follow up for a e2e test targeting this behaviour.

apps/frontend/src/routes/(voters)/(located)/questions/+layout.ts
kaljarv, line 1: Try to drop this.

apps/frontend/src/routes/(voters)/+layout.svelte
```diff
@@ -0,0 +1,142 @@
...
+  import { VoterNav } from '$lib/dynamic-components/navigation/voter/';
+  import Layout from '../Layout.svelte';
+  import MaintenancePage from '../MaintenancePage.svelte';
+  import type { Snippet } from 'svelte';
+
+  ////////////////////////////////////////////////////////////////////
+  // Init Voter Context
+  ////////////////////////////////////////////////////////////////////
+
+  // WR-04 (see phase 86.3 review): popupQueue is a stable instance reference per
```
kaljarv, line 36: Make sure that in the comment cleanup phase, extended prose like this is removed, perhaps completely. We can use this file as an example of what the code should look like afterwards.

apps/frontend/src/routes/(voters)/+layout.svelte
```diff
@@ -0,0 +1,142 @@
...
+  const appSettings = $derived(ctx.appSettings);
+  appType.set('voter');
+
+  ////////////////////////////////////////////////////////////////////
+  // Layout
+  ////////////////////////////////////////////////////////////////////
+
+  const { navigation, useTopBar } = getLayoutContext();
+
+  // see phase 86.3-01 wave A fix (cells #1 + #2): the top-bar overlay must
```
kaljarv, line 59: As well as this.

apps/frontend/src/routes/(voters)/+layout.svelte
```diff
@@ -0,0 +1,142 @@
...
+        help: help ? ('show' as const) : ('hide' as const)
+      }
+    });
+  });
+
+  ////////////////////////////////////////////////////////////////////
+  // Popup management
+  ////////////////////////////////////////////////////////////////////
+
+  // see phase 86.3-01 cell #3 (notifications.voterApp) REVERTED 2026-05-20 to the
```
kaljarv, line 92: Here we can leave one short line justifying onMount but no planning details.

apps/frontend/src/routes/(voters)/+layout.svelte
```diff
@@ -0,0 +1,142 @@
...
+  // downstream voter spec. Cells #1 + #2 (header.showFeedback / showHelp via
+  // reactive topBar $effect above) remain reactive — those changed in app
+  // settings UI and the topBar reactivity is the user-visible value.
+  //
+  // Cell #3 disposition is now REVERT-TO-ONMOUNT (downgraded from FIX-PASS),
+  // matching the baseline (see phase 77). The DataConsentPopup branch (below) stays
+  // in the same onMount, per the same small-fix constraint.
+  onMount(() => {
+    if (!appSettings.access.voterApp) return;
+    // Queue the voter-app notification popup (cell #3 — onMount one-shot).
```
kaljarv, line 109: Remove comment

apps/frontend/src/routes/(voters)/+layout.svelte
```diff
@@ -0,0 +1,142 @@
...
+    if (appSettings.notifications.voterApp?.show) {
+      popupQueue.push({
+        component: Notification,
+        props: { data: appSettings.notifications.voterApp }
+      });
+    }
+    // Ask for event tracking consent if we have no explicit answer
```
kaljarv, line 116: This can be kept, because the logic is a bit slow to read from the if clause

apps/frontend/src/routes/admin/login/+page.server.ts
```diff
@@ -0,0 +1,61 @@
...
+  default: async ({ request, locals }) => {
+    const data = await request.formData();
+    const email = data.get('email') as string;
+    const password = data.get('password') as string;
+    // Caller-controlled: see `loginRedirectTarget.ts`. Rejected values fall back to the app home.
+    const redirectTo = safeRedirectTarget(data.get('redirectTo') as string | null);
+
+    // Sign in directly via the Supabase server client from hooks.server.ts.
+    // This ensures session cookies are set on THIS response (not a nested API route response).
+    const { error } = await locals.supabase.auth.signInWithPassword({ email, password });
```
kaljarv, line 27: Add as a blocking follow-up task a way to make this supabase independent in routes. It should be handled by the SupabaseAdapter.

We can structure the abstracted model on supabase, though.

Also, add a source test for ensuring that no adapter-specifics find their way into routes, components or anywhere not especially allowed, such as the specific adapter's implementation and hand-picked locations.

apps/frontend/src/routes/admin/login/+page.server.ts
```diff
@@ -0,0 +1,61 @@
...
+    const { session, user } = await locals.safeGetSession();
+    if (!session || !user) {
+      logDebugError('Admin login: session not established after signIn');
+      return fail(500);
+    }
+
+    // Check user has an admin role via JWT claims.
+    const payload = JSON.parse(atob(session.access_token.split('.')[1]));
+    const userRoles: Array<{ role: string }> = payload.user_roles ?? [];
+    const isAdmin = userRoles.some((r) => ['project_admin', 'account_admin', 'super_admin'].includes(r.role));
```
kaljarv, line 43: This should be extracted to a /lib/auth or so utility for centralised permissions mapping.

apps/frontend/src/routes/api/auth/login/+server.ts
kaljarv, line 1: Consider this also as part of the login harmonisation.

apps/frontend/src/routes/api/auth/logout/+server.ts
```diff
@@ -1,18 +1,12 @@
...
  */
-export async function POST({ cookies }) {
-  cookies.delete(AUTH_TOKEN_KEY, {
-    httpOnly: true,
-    secure: true,
-    sameSite: 'strict',
-    path: '/'
-  });
+export async function POST({ locals }) {
+  await locals.supabase.auth.signOut();
```
kaljarv, line 10: And this as well.

apps/frontend/src/routes/api/candidate/preregister/+server.ts
```diff
@@ -0,0 +1,58 @@
...
+
+  if (!idToken) {
+    error(401, { message: 'ID token has expired.' });
+  }
+
+  // Call identity-callback Edge Function with the raw id_token.
+  // The Edge Function handles JWE decryption, JWT verification, user/candidate
+  // creation, and returns a magic link for session establishment.
+  try {
+    const { data, error: fnError } = await locals.supabase.functions.invoke('identity-callback', {
```
kaljarv, line 16: Check.

apps/frontend/src/routes/api/oidc/callback/+server.ts
```diff
@@ -0,0 +1,103 @@
...
+import type { RequestEvent } from '@sveltejs/kit';
+
+export async function GET({ url, cookies }: RequestEvent): Promise<never> {
+  const code = url.searchParams.get('code');
+  const returnedState = url.searchParams.get('state');
+  const errorParam = url.searchParams.get('error');
+
+  // Handle IdP errors (user canceled, access denied, etc.)
+  if (errorParam) {
+    throw redirect(303, '/candidate/preregister?error=' + encodeURIComponent(errorParam));
```
kaljarv, line 30: Get the return route from the centralised routes.

apps/frontend/src/routes/api/oidc/authorize/+server.ts
```diff
@@ -0,0 +1,54 @@
...
+      return error(400, { message: 'redirectUri is required' });
+    }
+
+    const provider = getActiveProvider();
+    const result = await provider.getAuthorizeUrl({ redirectUri, codeChallenge });
+
+    // Store state and nonce in httpOnly cookies for verification on callback.
+    // Idura returns these from JAR construction; Signicat uses client-side PKCE instead.
+    if (result.state) {
+      cookies.set('oidc_state', result.state, {
```
kaljarv, line 31: Move cookie names into a const so that they match between files.

Check all cookies written by the app and move those in the same const as well and place it in lib/cookies or so that we can make sure cookie names never overlap.

apps/frontend/src/routes/api/oidc/callback/+server.ts
```diff
@@ -0,0 +1,103 @@
...
+  }
+
+  // Validate authorization code is present
+  if (!code) {
+    throw redirect(303, '/candidate/preregister?error=missing_code');
```
kaljarv, line 35: Check whether we could use strict type for the error parameters.

apps/frontend/src/routes/candidate/(protected)/profile/+page.svelte
```diff
@@ -0,0 +1,366 @@
...
+
+  <!-- Editable data -->
+
+  <section>
+    <h2 class={subheadingClass}>{t('candidateApp.basicInfo.editableInfos.title')}</h2>
+
+    <div class="gap-md flex flex-col">
+      <!-- Image -->
+
+      <!--
```
kaljarv, line 281: NEVER add an element just for testid purposes. Put the testid in the parent and use that + a child selector in the test instead.

apps/frontend/src/routes/candidate/(protected)/+page.svelte
```diff
@@ -0,0 +1,165 @@
...
+  const appSettings = $derived(candCtx.appSettings);
+
+  ////////////////////////////////////////////////////////////////////
+  // Create action list
+  ////////////////////////////////////////////////////////////////////
+
+  // React to changes in language and stores
+  let nextAction = $derived.by(() => {
+    const username = userData.current?.candidate.firstName || '?';
+    if (candCtx.profileComplete) {
```
kaljarv, line 38: Rewrite this to be easier to read such that each prop gets a default at start or at the end and then we spefcify overrides per case.

Define the possible badges at the same time so that later below only the precomputed props are referenced in the if blocks.

apps/frontend/src/routes/candidate/auth/callback/+server.ts
```diff
@@ -0,0 +1,54 @@
+import { redirect } from '@sveltejs/kit';
+import type { EmailOtpType } from '@supabase/supabase-js';
+import type { RequestHandler } from './$types';
+
+/**
+ * Auth callback route for Supabase PKCE token exchange.
+ *
+ * Handles all Supabase auth redirects:
```
kaljarv, line 8: Check supabase specifics.

Also, we should move this to the generic /api routes because we'll be needing some of the same functions for admins as well.

apps/frontend/src/routes/candidate/auth/callback/+server.ts
```diff
@@ -0,0 +1,54 @@
...
+  const type = url.searchParams.get('type') as EmailOtpType | null;
+  const next = url.searchParams.get('next');
+  const lang = locals.currentLocale ?? 'en';
+
+  if (token_hash && type) {
+    const { error } = await locals.supabase.auth.verifyOtp({ token_hash, type });
+    if (!error) {
+      switch (type) {
+        case 'recovery':
+          redirect(303, `/${lang}/candidate/password-reset`);
```
kaljarv, line 31: Construct routes using buildRoute

apps/frontend/src/routes/candidate/auth/logout/+server.ts
kaljarv, line 1: Use the generic api route.

apps/frontend/src/routes/candidate/login/+page.server.ts
kaljarv, line 1: See earlier comment on admin login.

Also, extract the common logic from both to reduce duplication.

If this leaves the generic /api login route unused, delete it.

apps/frontend/src/routes/candidate/preregister/(authenticated)/elections/+page.svelte
kaljarv, line 1: Add a follow-up task: harmonise election and constituency selection logic with the Voter App (mainly startFromConstituencyGroup option).

apps/frontend/src/routes/candidate/preregister/+layout.server.ts
```diff
@@ -1,10 +1,17 @@
...
 import { redirect } from '@sveltejs/kit';
 import { getIdTokenClaims } from '$lib/api/utils/auth/getIdTokenClaims';
 import { buildRoute } from '$lib/utils/route';
 
 export async function load({ cookies, locals }) {
-  if (!staticSettings.preRegistration.enabled) {
+  // `preRegistration.enabled` is a DynamicSetting (backend/per-instance controlled,
+  // stored in the `app_settings` JSONB row) — read it server-side via the request's
+  // Supabase client rather than from build-time StaticSettings.
+  const { data: appSettingsRow } = await locals.supabase.from('app_settings').select('settings').limit(1).maybeSingle();
```
kaljarv, line 9: Extract this to the adapter and use the typing provided with no ad hoc casts.

apps/frontend/src/routes/+layout.svelte
```diff
@@ -0,0 +1,284 @@
...
+
+  // popupItem reactivity is handled inline at the template tail via
+  // popupQueue.current + {@const Component = item.component}
+
+  const fontUrl =
+    staticSettings.font?.url ?? 'https://fonts.googleapis.com/css2?family=Inter:wght@400;700&display=swap';
+</script>
+
+<svelte:head>
+  <title>{underMaintenance ? t('maintenance.title') : t('dynamic.appName')}</title>
```
kaljarv, line 212: {t('dynamic.appName')}
{#if underMaintenance}
  – {t('maintenance.title')}
{/if}

apps/frontend/src/routes/+layout.svelte
```diff
@@ -0,0 +1,284 @@
...
+  <meta
+    name="theme-color"
+    content={staticSettings?.colors?.light?.['base-300'] ?? '#d1ebee'}
```
kaljarv, line 215: Remove the defaults from theme-colors.

apps/frontend/src/routes/Banner.svelte
```diff
@@ -0,0 +1,116 @@
+<!--@component
+
+# Banner component
+
+Contains the secondary action buttons in the header.
+
+### TODO
+
+Allow layouts to insert arbitrary content in the header and make this a static component [Svelte 5].
```
kaljarv, line 9: Add this is as a follow up task.

apps/frontend/src/routes/Header.svelte
```diff
@@ -14,30 +14,37 @@ Accesses `AppContext` and renders the dynamic `Banner` component.
...
+  // `darkMode` is a stable `{ readonly current }` rune handle from AppContext.
+  // `appSettings` is a reactive accessor (see phase 113 flatten) — read via
+  // `ctx.appSettings`, never destructure (the alias below tracks it).
+  const ctx = getAppContext();
+  const { darkMode, t } = ctx;
+  const appSettings = $derived(ctx.appSettings);
+
+  const { navigationSettings, progress, topBarSettings } = getLayoutContext();
+
+  const bgColor = $derived.by(() => {
```
kaljarv, line 44: Add as a follow up task, refactoring the header style settings.

apps/frontend/src/routes/loginRedirectTarget.ts
kaljarv, line 1: Move this to lib/routes

apps/frontend/src/hooks.server.ts
```diff
@@ -1,133 +1,88 @@
...
-
-  //////////////////////////////////////////////////////////////////////////
-  // 3. Redirect if the locale param is not the same as the served locale
-  //////////////////////////////////////////////////////////////////////////
+/**
+ * Supabase session handler.
+ * Creates a per-request server client and attaches it (plus safeGetSession) to event.locals.
+ * Runs FIRST so all subsequent handlers can use event.locals.supabase.
+ */
+const supabaseHandle: Handle = async ({ event, resolve }) => {
```
kaljarv, line 17: Paramaterise this dependent on the adapter configuration and rename to dataAdapterHandle if possible.

apps/frontend/src/hooks.server.ts
```diff
@@ -1,133 +1,88 @@
...
-    const token = event.cookies.get(AUTH_TOKEN_KEY);
-
-    if (token && pathname.endsWith('candidate/login')) {
-      debug('Route: REDIRECT to home page');
-      return new Response(undefined, {
-        headers: { location: `/${servedLocale}/candidate` },
-        status: 303
-      });
+  // Handle candidate auth redirects
+  if (pathname.includes('/candidate')) {
```
kaljarv, line 69: The check must be more restrictive because candidate might be used on a subpath.

Also, move all path definitions to the centralised routes locus. Define the (protected) pattern there as well and add tests to enforce consistency.

kaljarv (commented):
Checked.

PR #871, opened 2026-08-17
9/12 feat: add the Paraglide message catalogues for every supported locale

kaljarv:
# 9/12 · `feat: add the Paraglide message catalogues for every supported locale`

**Stack position: 9 of 12.** Base: `ship/v0.2-akita-07-frontend-routes`. Head: `ship/v0.2-akita-08-i18n-messages`.
**330 files changed, 8,986 insertions, 0 deletions** — and unusually for this stack, GitHub will show
you exactly that. Rename detection changes nothing here: every one of the 330 files is an addition
(`330 A`, no `D` and no `M`), so there is no delete-plus-add pair for GitHub to pair up.
`git show --shortstat` and `git show --shortstat --no-renames` return the identical triple. Both
numbers are stated because elsewhere in this stack they differ sharply — PR 8 renders as 165 files
against a measured 214 — and an unreconciled pair is what makes a reviewer distrust a header line.

**This PR is one shape repeated 329 times.** The Paraglide message catalogue: 7 locales × 47
namespace files, plus one README. There is no logic in it. What is worth your attention is not any
individual file but whether the 7 locales agree with each other — and that was measured exhaustively
rather than sampled, because the failure mode is silent.

---

   zero files.

not asserted: every off-diagonal cell of the pairwise overlap matrix is `0`, and the eleven cut slices
plus the untouched remainder reproduce the development branch's tree hash exactly — most recently
`40f5d20c5`, measured after slices 09 and 10 were cut. The partition arithmetic closes at a gap of
`0`: 252 + 97 + 119 + 162 + 195 + 533 + 214 + 330 + 152 + 129 + 2,321 = **4,504**, the comparable file
total. And the remainder decomposes independently of that measurement: after slice 10 the only files
left are `.planning/`, `.claude/`, `.agents/` and `CLAUDE.md` — **zero** files outside slice 11's own
pathspec.

## What you are looking at

`apps/frontend/messages/` — the message catalogue Paraglide compiles into typed message functions.

| Area | Files | What it is |
| --- | ---: | --- |
| `messages/{locale}/*.json` | 329 | 7 locales × 47 namespace files. `en` is the base locale; the others are `fi`, `sv`, `da`, `et`, `fr`, `lb`. |
| `messages/README.md` | 1 | Written during this review pass: what the catalogue is, how a key becomes a callable message, and the parity checks below as runnable scripts. |

The 47 namespaces are declared in `apps/frontend/project.inlang/settings.json`, which ships in PR 11
(the configuration slice) — so a reviewer of *this* PR cannot see the file that decides which of these
files are compiled. That is stated rather than left to be discovered, and the consistency between the
two was measured here: **the 47 `pathPattern` entries and the 47 files on disk have a symmetric
difference of 0, in both directions.** No file is uncompiled and no pattern entry lacks a file.

**One property of this slice is worth stating because most slices in this stack do not have it: the
diff is the entire surface.** Elsewhere a file can be byte-identical across the repository layout move
that PR 1 performs, which means PR 1 renders it as a rename and no later PR's diff contains it —
reviewed by nobody. Under this slice's paths there is **no such file**, established two independent
ways: files tracked at the head under the pathspec minus files present in the diff is `330 − 330 = 0`,
and a blob-by-blob comparison against the pre-move tree finds **0** identical files. Nothing here is
invisible.

Backend, Supabase Adapter, Edge Functions) do not apply — this slice touches none of their paths.

### Locale parity, measured over every locale, every file and every key

This is the check the slice exists to answer, and the answer is exact rather than a sample.

| Measurement | Value |
| --- | ---: |
| locale directories on disk | **7** |
| locales declared in `project.inlang/settings.json` | **7**, identical set, base `en` |
| files per locale — and the union of file names across all 7 | **47 = 47** — no locale missing a namespace, none with an extra |
| `pathPattern` entries vs files on disk | **47 = 47**, symmetric difference **0** both ways |
| fully-qualified keys per locale | **598** in every one of the 7 |
| union of all locales' key sets | **598** |
| intersection of all locales' key sets | **598** |
| **symmetric difference (union − intersection)** | **0** |

**No key is present in one locale and missing in another.** The scripts that produce these numbers are
in `messages/README.md` in this PR, and were executed as written before being committed.

**Why this had to be measured directly rather than left to a test.** The translation wrapper returns
**the key string itself** when it cannot resolve a message, so a locale missing a key renders a raw
dotted identifier like `candidateApp.settings.title` into the page instead of failing. There is a
check for that — but its reach is the accessibility scan's route table, which resolves to **5 URLs**,
none of them in the candidate or admin apps. A parity defect in a Finnish candidate-app string is
outside that check by construction. The compile-time key union does not close the gap either: it is
generated from the *legacy* catalogue that ships in PR 7, not from these files.

### Placeholder parity — the second measurement, and it found the slice's one defect

A key can be present in all seven locales and still be broken, because `{name}` refers to a **declared
input** and the declared names are part of the message's callable signature. Compared across all 598
keys × 7 locales, recursing into plural and selector bodies — **147 message bodies are arrays rather
than strings, and a check that skipped them would have reported a clean result and been wrong**:

| | Before | After |
| --- | ---: | ---: |
| placeholder-set divergences against `en` | **2** | **1** |

- **Fixed before this PR opened.** `sv/components.json` → `components.video.timeLeft` declared its
  inputs as `minuter` / `sekunder` and interpolated `{minuter}` / `{sekunder}`, where all six other
  locales declare `minutes` / `seconds`. The Swedish variant therefore had a **different callable
  signature** from its siblings and could not receive what a caller passes. Only the four
  declarations, two selector names and two interpolations were renamed — **the Swedish text is
  untouched**, so no translation decision was taken. Severity stated rather than inflated: the key has
  **zero call sites** in the repository today, so nothing renders it; it was fixed because the repair
  is mechanical and lossless. It is also inherited rather than introduced — the pre-Paraglide
  catalogue carries the same names.
- **One divergence remains, and it is deliberate.** `questions.intro.start` is
  `Answer {numQuestions} Questions` in `en` and interpolates `numQuestions` in five other locales;
  `sv` reads `Börja svara på frågor` with no placeholder. That one **is** live — it is the
  questions-intro call to action — but the remedy is Swedish copy, which is a publisher's decision
  and not a reviewer's or an implementer's.

Two further measurements, neither of which is a defect: **12 empty-string values**, of which 7 are one
key that is empty in all seven locales by design (a publisher-overridable default whose render site is
guarded on truthiness) and 5 are a shared suffix; and **31 values identical to `en` in all six other
locales**, of which 24 are emoji, 7 are the language endonyms in `lang.json` (`Suomi`, `Svenska`, …,
correctly untranslated), and one is that same empty default.

## Why no checks run here

`ship/v0.2-akita-07-frontend-routes`, a sibling branch, so **no checks run — and that is correct

**The reason is the base branch, not the file types in the diff.** The workflow does carry a
`paths-ignore` for `**.md`, and it is irrelevant here twice over: this PR is 329 `.json` files and one
`.md`, and `paths-ignore` only ever applies to a run the trigger already selected.

**Do not read "no checks" as "the jobs exist and were skipped."** The workflow file at this branch's
head is still `main`'s — byte-identical, blob `c2fdcedb2` — and it defines exactly three jobs:
*exist* here. The branch-tip workflow, with its newer Yarn and its four additional jobs
(`skill-drift-check`, `supabase-tests`, `dev-seed-integration`, `e2e-visual`), arrives with PR 11.

reason that has nothing to do with this PR. Observed on PR 1's run, and re-verified against that run
rather than copied: all three fail at step **`Setup Yarn 4.6`** with

```
```

never reached and is reported `skipped`. The cause is that the root `package.json` at this head still
declares `frontend`, `docs` and `backend/vaa-strapi` as workspaces — verified at this exact commit —
while PR 1 moved two of them under `apps/` and PR 2 deleted the third. Those 412 packages dropping out
of the resolution are the dependency closure of the moved-away workspaces. The root `package.json` and
`yarn.lock` are corrected in `ship/v0.2-akita-10-root-config` (PR 11), so **every stack state from
PR 1 up to and including PR 10 — ten branches — cannot `yarn install`.** That is the stack contract's
point (1) working as designed.

**What was actually run, locally, against the development branch tip these slices are cut from:**
`yarn build` 14/14 packages, `yarn test:unit` **1,522 tests across 149 files**, `yarn lint:check`
**0 errors**, each under a forced, uncached run and each identical to the pre-review baseline. **The
43 end-to-end specs were not run**; that suite needs a dev server and a seeded local database, and
this project's own rule is that a did-not-run E2E test counts as a failure rather than a pass, so **no
green suite is claimed for this PR.** The full-suite run against the post-review tip is the stack's
collective proof.

**330 files** is just over the ~300-file point at which the Files-changed tab starts truncating, so
you may need to load the remainder. At 8,986 changed lines it is comfortably inside the 20,000-line
limit, so no individual diff collapses.

The truncation costs you little here, because the 329 JSON files are one shape: if you have read
`en/` and one other locale, you have read the pattern. **The parity tables above are the part that
cannot be obtained by scrolling**, which is why they are stated as counts with their commands rather
than as an impression.

## The stack

`main`.

| **9** | **`ship/v0.2-akita-08-i18n-messages`** ← *you are here* | add the Paraglide message catalogues for every supported locale | **330** | **8986** | **0** |
| 10 | `ship/v0.2-akita-09-docs` | update the project documentation — the docs site, the root README and roadmap, and the key-generation guide | 152 | 777 | 347 |
| 11 | `ship/v0.2-akita-10-root-config` | update the monorepo and frontend-app build, lint, CI and deployment plumbing | 129 | 8662 | 27267 |
| 12 | `ship/v0.2-akita-11-planning` | add the v0.2 planning record and agent configuration | 2321 | 878943 | 104 |

Line counts are measured over each slice's own commit with rename detection off. Row 1's is the
rename-aware measurement, which is what a reviewer actually sees — GitHub's own API reports
`files=1316, +0, -0` with 1,316 `renamed` entries. Rows 1–11 are measured on cut slices; row 12 is
measured from the remainder and moves slightly as each plan writes its own record.

## Full evidence

Every verdict above, with file-and-line citations — including the per-locale key and placeholder
tables and the full list of empty and identical values — is in
**"Slice 08 — `ship/v0.2-akita-08-i18n-messages`"**.

[151-DISPOSITION.md § Slice 08](https://github.com/OpenVAA/voting-advice-application/blob/ship/v0.2-akita-11-planning/.planning/phases/151-ship-v0-2-akita-review-stack/151-DISPOSITION.md#slice-08--shipv02-akita-08-i18n-messages)


kaljarv (commented):
Bypassed as non-critical.

PR #872, opened 2026-08-17
10/12 docs: update the project documentation - the docs site, the root README and roadmap, and the key-generation guide

kaljarv:
# 10/12 · `docs: update the project documentation - the docs site, the root README and roadmap, and the key-generation guide`

**Stack position: 10 of 12.** Base: `ship/v0.2-akita-08-i18n-messages`. Head: `ship/v0.2-akita-09-docs`.
**152 files changed, 777 insertions, 347 deletions** — and GitHub will show you exactly that. Rename
detection changes nothing here: the slice is `2 A / 150 M`, with no deletion that could pair with an
addition, so `git show --shortstat` and `--no-renames` return the identical triple. Both numbers are
stated because elsewhere in this stack they differ sharply — PR 8 renders as 165 files against a
measured 214 — and an unreconciled pair is what makes a reviewer distrust a header line.

**This PR is all of the project's non-planning documentation, and it is deliberately one commit.**
The restructure this stack implements requires that all documentation outside the planning record
collapse to a single `docs:` commit; this slice's pathspec and that requirement are the same set by
construction, not by a squash that happened to gather them.

**Most of what you are reading is a repair, not new prose.** The v1.1 repository layout move took
`frontend/` and `docs/` under `apps/`, and the documentation was never fully updated for it. This PR
is where that is settled.

---

   zero files.

slice's diff, in its final state. Verified, not asserted: every off-diagonal cell of the pairwise
overlap matrix is `0`, and the eleven cut slices plus the untouched remainder reproduce the
development branch's tree hash `40f5d20c5` exactly. The partition arithmetic closes at a gap of `0`:
252 + 97 + 119 + 162 + 195 + 533 + 214 + 330 + 152 + 129 + 2,321 = **4,504**, the comparable file
total. And the remainder decomposes independently of that: after the next PR, the only files left are
`.planning/`, `.claude/`, `.agents/` and `CLAUDE.md` — **zero** files outside the final PR's own
pathspec.

## What you are looking at

`apps/docs` (the documentation site), the root `ROADMAP.md`, the root `README.md`, and
`docs/key-generation.md`.

| --- | ---: | --- |
| `apps/docs/src/routes/(content)/developers-guide/**` | ~30 | Guide content: legacy-backend notices, corrected commands, corrected source links. |
| `apps/docs/src/routes/(content)/developers-guide/frontend/components/generated/**` | 98 | Generated component reference pages. **Source-link repair only** — one line per page. |
| `apps/docs/src/routes/(content)/publishers-guide/**` | 2 | The app-settings page and one corrected source link. |
| `apps/docs/{package.json, tailwind.config.mjs, tsconfig.json, scripts/, src/lib/}` | 7 | Build and navigation configuration for the docs site itself. |
| `apps/docs/src/routes/+page.svelte` | 1 | The site landing page. |
| `ROADMAP.md` | 1 | **New.** The roadmap this review restructure itself implements. |
| `docs/key-generation.md` | 1 | **New.** How to generate the signing keys the bank-authentication flow needs. |
| `README.md` | 1 | The repository front page — one line, and it is a real bug fix. See below. |

**Two files are added; the other 150 are modifications.** There is no deletion in this PR.

### 1. The front page's image was broken, and nobody could see it

`README.md:12` rendered ``. **That path
does not exist.** The v1.1 layout move took the image to `apps/docs/static/images/`, and the README
was not updated with it — so the first thing a visitor to the repository sees has been a broken image.

It stayed invisible for a structural reason worth naming, because it applies to more than this file:
`README.md` is **byte-identical at both ends of this stack**, so PR 1 renders it as a rename and no
later PR's diff contained it. A reviewer working through twelve PRs would never have been shown it.
It is in this diff only because the slice's file scope was deliberately widened to claim it.

### 2. 273 source links pointed at a path that no longer exists

Across **117 files**, links of the form
`github.com/OpenVAA/voting-advice-application/blob/main/frontend/…` pointed at the pre-move location.
**240 of the 252 distinct targets resolve** once the `apps/` prefix is added; that was verified target
by target rather than assumed, with `git cat-file -e` against each.

**The 12 that still do not resolve are named rather than hidden** — they point at files v0.2 deleted
(the Strapi data adapter, the store-era state modules, three components that were rewritten). All 12
are in hand-authored pages, not in the 98 generated ones, and each needs a semantic re-target rather
than a path prefix. They are recorded as an open item with their carrying pages listed.

Worth knowing about the timing: these links resolve on `main` **today** and stop resolving the moment
this work merges, because v0.2 both moves the tree and deletes those files. So the repair is correct
for the state this stack produces, and you cannot check it by clicking the *old* link.

### 3. Fifteen pages cited `apps/strapi/`, a path that has never existed

v0.2 rewrote the Strapi source links in the backend guides from `backend/vaa-strapi/…` to
`apps/strapi/…` — **48 occurrences**. The Strapi backend was *deleted* by v0.2, not moved to
`apps/strapi`: that directory has never existed in this repository, at either end of this stack. It is
the same defect that was found and fixed in `.prettierignore`, which shipped six ignore rules for the
same imaginary path, and it comes from the same blanket rewrite.

They are restored to `backend/vaa-strapi/…`, the path that did exist. **22 of the 30 distinct targets
resolve at `origin/main`; the other 8 were already dead there**, before any of this work — recorded,
not chased. Every one of these pages carries v0.2's own banner saying it documents the legacy backend.

### 4. Four guides documented commands that no longer exist

A guide naming a removed script is worse than a guide saying nothing, so each was replaced with the
command as it exists now — checked against the root `package.json`'s actual script list, not against
the pre-harmonisation names:

- `yarn dev:down` and `yarn dev:stop` (3 sites) → `yarn db:stop` and `yarn db:reset`. The Docker dev
  stack those belonged to was replaced by a local Supabase instance.
- `yarn build:app-shared` → `yarn build --filter=@openvaa/app-shared`.
- `yarn prod` → `docker compose -f docker-compose.dev.yml up --build`, which is the command that
  compose file's own header documents. The same paragraph promised a backend container that no longer
  exists.

The paragraph describing `yarn dev` as "development Docker images for the entire stack (frontend,
backend and DB)" is corrected too: it now describes what the script actually does.

### 5. The workspace inventory listed two workspaces that are gone and missed four that exist

`developers-guide/app-and-repo-structure` enumerated 13 workspaces, including `@openvaa/strapi` and
`@openvaa/strapi-admin-tools` (deleted by this body of work) and omitting `@openvaa/supabase`,
`@openvaa/dev-seed`, `@openvaa/dev-tools` and `@openvaa/supabase-types`. The repository has **15**.

Verified by set comparison against `yarn workspaces list --json` rather than by reading: **the
symmetric difference is empty in both directions** — nothing listed that is not a workspace, no
workspace unlisted. The page's "parts of this page reference the legacy Strapi backend" banner is
removed with the two dead entries, because after they go the page has no legacy content left, and a
banner that no longer applies is itself a documentation defect.

### 6. The landing page's external link had no `rel`, and the page was unformatted

`src/routes/+page.svelte` gained a block from `main` through an integration merge. Its
`target="_blank"` anchor carried no `rel="noopener noreferrer"`, while the sibling component in the
same app does. Modern browsers imply `noopener`, so the severity is low — the argument is the
codebase's own convention, and the block is produced by no v0.2 commit, so no other slice's diff would
ever have shown it to anyone.

The same merged block was indented one space where its siblings use two, which made the docs
workspace's **own** `format:check` fail. That gate had never run during this review: the repository's
root `format:check` chains the two checks with `&&` and short-circuits on two pre-existing, cosmetic
failures elsewhere, so the docs half was never reached. It is now green.

Backend, Supabase Adapter, Edge Functions) do not apply.

Where an item is enumerable it is reported as a count with the command behind it:

| Check | Checked over | Result |
| --- | ---: | ---: |
| `any` in type position | 152 files | **0** |
| `@ts-expect-error` / `@ts-ignore` | 152 files | **0 / 0** |
| `{@html}` occurrences | whole docs app | **0** |
| secret-shaped literals | 152 files | **0** |
| `` elements / of those with `alt` | whole docs app | **13 / 13** |
| positive `tabindex` | whole docs app | **0** |
| non-semantic elements with a click handler | whole docs app | **0** |
| `autofocus` in code | whole docs app | **0** |
| `svelte-ignore` suppressions | whole docs app | **0** |
| internal links, of which broken | 191 files, 166 links | **0 broken** |
| `svelte/store` imports | whole docs app | **0** |

**Three limits on that, stated because the alternative is a claim:**

- **No automated accessibility gate reaches this site at all.** The end-to-end accessibility scan's
  route table is the *application's* five URLs. The docs site has no scan, no end-to-end spec, and its
  `test:unit` is `vitest run --passWithNoTests` — an empty pass. The table above is a static sweep over
  an enumerated class list, not a gate result, and **keyboard and screen-reader use was not
  traversed**.
- **`yarn lint:check` does not reach this workspace either.** It defines `lint:local` and `lint:full`
  but no `lint` task, so the monorepo's lint runner skips it silently. The `any` result above is a
  grep, not a linter.
- **The 98 generated pages cannot currently be regenerated from the documented entry points.** Five of
  the seven root `docs:*` scripts call workspace scripts that do not exist — `yarn docs:generate`
  answers `Couldn't find a script named "generate"`. That is pre-existing, identical on `main`, and
  it is also the mechanism behind the stale links in item 2. It is recorded rather than fixed, because
  two of the five have no correct target at all.

## About the next PR in the stack

The PR above this one (11 of 12, the configuration and plumbing slice) **deletes 89 files of
Capacitor native scaffolding** — `apps/frontend/{android,ios}` — together with a dead jest config.
That is flagged here because it is a removal a reviewer should be able to object to, and because the
evidence for it is worth stating once in a body rather than only in a commit message.

Every in-repo signal says the scaffold is dead: `@capacitor/*` appears in **no** `package.json`,
`yarn.lock` has **zero** capacitor entries, no source file imports it, and the only `@capacitor`
strings anywhere in the tree were inside the scaffold itself. `capacitor.config.ts` is deleted by
v0.2, so a reviewer of that PR would otherwise see the config go and reasonably assume the removal was
already complete.

**What no in-repo measurement can see is an external build.** If a Capacitor or app-store pipeline
outside this repository consumes that scaffold, the removal breaks it. The maintainers accepted that
risk explicitly; it is stated here so that anyone who knows of such a pipeline can object on the
strength of this text alone.

## Why no checks run here

`ship/v0.2-akita-08-i18n-messages`, a sibling branch, so **no checks run — and that is correct

**The reason is the base branch, not the file types in the diff.** The workflow does carry a
`paths-ignore` for `**.md`, and it does not apply here: 7 of this slice's files are `.svelte`, `.ts`,
`.mjs` and `.json`, and in any case `paths-ignore` only filters a run the trigger already selected.

**Do not read "no checks" as "the jobs exist and were skipped."** The workflow file at this branch's
head is still `main`'s — byte-identical, blob `c2fdcedb2`, verified at this commit — and it defines
exactly three jobs: `frontend-and-shared-module-validation`, `backend-validation` and `e2e-tests`. The
branch-tip workflow, with Yarn 4.13 and four additional jobs (`skill-drift-check`, `supabase-tests`,
`dev-seed-integration`, `e2e-visual`), **arrives with the next PR** — verified by comparing the file
at both heads, not assumed.

Had the three that are present run, they would have failed at the same place PR 1 fails, for a reason
unrelated to this PR. Observed on PR 1's run and re-verified against it: all three fail at step
**`Setup Yarn 4.6`** with

```
```

`threeal/setup-yarn-action@v2` performs the install itself as part of its caching, so the workflow's
own `Install all dependencies` step is never reached and is reported `skipped`. The cause is that the
root `package.json` at this head still declares `frontend`, `docs` and `backend/vaa-strapi` as
workspaces — read at this exact commit — while PR 1 moved two and PR 2 deleted the third. **The next
PR is where that stops being true**: it carries the corrected `package.json` and `yarn.lock`, so every
stack state from PR 1 up to and including this one — ten branches — cannot `yarn install`.

**What was actually run, locally, against the development branch tip these slices are cut from, after
this PR's fixes landed:** `yarn build` 14/14 packages, `yarn test:unit` **1,522 tests across 149
files**, `yarn lint:check` **0 errors / 20 warnings**, `yarn format:check` failing on exactly the two
pre-existing cosmetic files it failed on before — each under a forced, uncached run and each identical
to the pre-review baseline. Additionally, and specific to this slice: the docs site **builds**, its
link validator reports **0 broken internal links**, and its own `format:check` is **green**. **The 43
end-to-end specs were not run**; that suite needs a dev server and a seeded local database, and this
project's rule is that a did-not-run E2E test counts as a failure rather than a pass, so **no green
suite is claimed for this PR.**

**152 files and 1,124 changed lines — comfortably inside both budgets** (the ~300-file limit before
the Files-changed tab truncates, and the 20,000-line limit before diffs collapse). The tab will render
this PR in full.

A reading order, because 98 of the 152 files are one mechanical line each: start with `README.md`,
then `ROADMAP.md` and `docs/key-generation.md` (the two additions), then the hand-authored guide pages
under `developers-guide/`, and leave `developers-guide/frontend/components/generated/**` until last —
every one of those is the same single-line link repair.

## The stack

`main`.

| **10** | **`ship/v0.2-akita-09-docs`** ← *you are here* | update the project documentation — the docs site, the root README and roadmap, and the key-generation guide | **152** | **777** | **347** |
| 11 | `ship/v0.2-akita-10-root-config` | update the monorepo and frontend-app build, lint, CI and deployment plumbing | 129 | 8662 | 27267 |
| 12 | `ship/v0.2-akita-11-planning` | add the v0.2 planning record and agent configuration | 2321 | 878943 | 104 |

Line counts are measured over each slice's own commit with rename detection off. Row 1's is the
rename-aware measurement, which is what a reviewer actually sees — GitHub's own API reports
`files=1316, +0, -0` with 1,316 `renamed` entries. Rows 1–11 are measured on cut slices; row 12 is
measured from the remainder and moves slightly as each plan writes its own record.

## Full evidence

Every verdict above, with file-and-line citations — including the reconciled list of every stale
documentation path in the repository with a terminal verdict for each, and the named list of the 12
links that still do not resolve — is in
**"Slices 09 and 10 — `-09-docs` and `-10-root-config`"**.

[151-DISPOSITION.md § Slices 09 and 10](https://github.com/OpenVAA/voting-advice-application/blob/ship/v0.2-akita-11-planning/.planning/phases/151-ship-v0-2-akita-review-stack/151-DISPOSITION.md#slices-09-and-10--09-docs-and--10-root-config)


kaljarv (commented):
Bypassed as non-critical.

PR #874, opened 2026-08-17
12/12 docs[planning]: add the v0.2 planning record and agent configuration

kaljarv:
**Stack position: 12 of 12 — the top.** Base: `ship/v0.2-akita-10-root-config`. Head: `ship/v0.2-akita-11-planning`.
**2,329 files changed, +882,299 / −104.**

> ## You can approve this PR without reading it. That is the design, not a concession.
>
> This is the planning record and agent configuration for the v0.2 "Akita" body of work. It exists as
> its own PR for exactly one reason: **to keep ~882,000 lines of planning artifacts out of the other
> eleven PRs' diffs.** Every one of those eleven is readable because this one absorbed the noise.
>
> It contains **no application code**. Nothing here is built, bundled, imported, or served — measured,
> not asserted: **0 of its 2,330 shipped files lie under `apps/` or `packages/`**, no `tsconfig`, `vite`,
> `svelte`, `vitest` or `turbo` config references any path in it, and `git grep` finds no import of it
> from product code.

---

This is one PR in a **review-only stack of twelve**, cut from the v0.2 body of work so that it can
actually be read. Four things follow, all deliberate:

1. **Individual PRs need not build, lint, or pass tests. Only the whole matters.** The stack slices one
   large final state into twelve diffs of similar nature. Intermediate states are not required to be
   working states.
   slices are a re-partition of the same tree, not a re-implementation.
   development branch or the stack produces the identical tree.
4. **Review comments need not be implemented inside the stack.** Fixes can land on a separate branch.

**No file is shown to you twice in two different content versions.**

## The identity proof — verify it yourself rather than trust it

This PR is the top of the stack, so the whole stack's central claim is checkable from here. Two
independent checks, computed by different code paths (diff machinery vs object hashing):

```bash
git diff feat-gsd-roadmap ship/v0.2-akita-11-planning          # prints nothing
git rev-parse feat-gsd-roadmap^{tree} ship/v0.2-akita-11-planning^{tree}   # prints one SHA, twice
```

Recorded result, at the cut proven in `.planning/phases/151-ship-v0-2-akita-review-stack/151-BYTE-IDENTITY-PROOF.md`:

| | |
|---|---|
| changed files | **0** |
| target tree (`feat-gsd-roadmap`) | **`c967a8457783764f9c9e9bbd9f5434cfe02f2f88`** |
| stack tip tree | **`c967a8457783764f9c9e9bbd9f5434cfe02f2f88`** |
| catch-all slice (the tripwire) | **`files=0`** |

**A caveat stated rather than hidden.** This PR's published head is a re-cut taken *after* the proof
above, because it necessarily contains this very PR body — **every `.planning/` file rides this slice
by pathspec**, so the record of the work is always one step behind the work. The closing summary of
the plan that opened this PR lands after it too, for the same structural reason. The identity is
re-verifiable at any tip with the one-liners above; the recorded pair is the measurement at the cut
the proof documents.

**A green identity check is not by itself proof of an honest split** — a non-empty catch-all slice
would launder a partition bug while both checks still passed (measured live during research: two
broken slices, 472 files absorbed, tree hash still matching). That is why the catch-all is asserted
empty separately, and why the real evidence is the per-slice prediction checks made while a wrong
answer was still catchable.

## The counts, both ways, reconciled

| measurement | value |
|---|---|
| **GitHub's own numbers** (API, this PR) | **2,329 files, +882,299 / −104** |
| measured `--no-renames` at the published head | **2,329 files, +882,299 / −104** |
| measured **with** rename detection (`-M`) | **2,329 files, +882,299 / −104** |
| status taxonomy | **2,327 `A`, 2 `M`, 0 `D`** |

**All three agree exactly, and the reason is worth stating** because for other slices in this stack
they do not: rename detection needs a deletion to pair an addition with, and this slice has **zero
deletions**. So there is no rename for GitHub to detect and no divergence to reconcile.

**One reconciliation that is worth being explicit about**, because the phase record quotes a smaller
figure. This body was first written against a cut measuring **2,328 files / +881,890**. The published
head is **2,329 / +882,299**. The difference is exactly one file — **this PR body itself**
(`pr-bodies/11.md`), which by construction did not exist when the earlier measurement was taken. The
other three files that changed in the same interval were modifications, not additions, so they moved
the line counts without moving the file count. Earlier figures in the planning record are correct
snapshots of the cut they were taken at; **the numbers above are the published head.**

Note also that the **shipped tree is 2,330 files while the diff shows 2,329**. The extra file is
`.claude/settings.json`, which is byte-identical at `main` and therefore appears in **no** slice's
diff anywhere in the stack. It is named here rather than left to be discovered.

## What is actually in here, if you want a route in

| path | files | what it is |
|---|---:|---|
| `.planning/` | 2,226 | The GSD planning record: per-phase context, research, plans, summaries and evidence captures across every milestone from v1.0 to v2.15. ~1,971 markdown files plus test-run evidence (`.json`, `.sha256`, `.txt`) and two Playwright trace archives. |
| `.claude/` | 101 | Agent configuration: 7 domain skills (`architect`, `components`, `data`, `database`, `filters`, `matching`, plus the spike-findings package), their reference docs, 16 frozen spike source directories, and `scripts/audit-skill-drift.sh`. |
| `CLAUDE.md` | 1 | The repository's agent instructions. **+243 / −104 — genuinely worth reading**, and one of only two files here that a reviewer should. |
| `.agents/` | 1 | `code-review-checklist.md`, **+38**. The other one worth reading. |

**If you read exactly two files, read `CLAUDE.md` and `.agents/code-review-checklist.md`.** They are
live repository documentation that shapes how everyone works in this repo. Everything else is a record
of how the work got done, valuable as history and as provenance, and not something to review line by
line.

## It was security-scanned before publication — the one control this PR does need

Nobody reads this diff, so an exposed credential in it would be found by nobody. It was therefore
scanned exhaustively *before* the branch was ever pushed. Full record:
`.planning/phases/151-ship-v0-2-akita-review-stack/151-SECRET-SCAN.md`.

| | |
|---|---|
| verdict | **`pass-after-remediation`** |
| **live findings** | **0** |
| findings | **15 total — 0 live, 1 remediated, 8 accepted, 6 false positive** |
| corpus | **2,325 files**, a `git archive` of the slice's pathspec — the *publication surface*, not the diff |
| methods | **TruffleHog 3.95.2** (all detectors) **plus** an independent 24-rule sweep **plus** targeted project-shape greps |

Three things that make this more than a formality, each of which changed the result:

- **Scanning the publication surface rather than the diff** caught `.claude/settings.json`, which ships
  here but appears in no diff at all.
- **One scanner is not a scan.** TruffleHog found 2 distinct JWTs; the independent sweep found **4**,
  adding Supabase demo `anon` and `service_role` tokens that no detector flagged. Both proved to be
  published public constants — established by **recomputing their HMAC** against Supabase's documented
  local-development secret, not by how they looked.
- **A working-tree grep is not a census when the content includes archives.** The one remediated
  credential existed in **32 places, 28 of them inside Playwright trace archives**; a working-tree grep
  reported 4. Both archives were rebuilt member-by-member with an equal-length replacement so the trace
  JSON and its recorded offsets stay valid.

Also measured: **zero** cloud hostnames (`*.supabase.co/.in/.net`) across all 2,325 files, and **zero**
JWK private key components.

## One thing the merger of this stack needs to know — F-88

**`skill-drift-check` will be red on the first PR opened against `main` after this stack merges, and
no PR in this stack shows it.**

The job arrives with slice 10 and the script it runs (`.claude/scripts/audit-skill-drift.sh`) arrives
with this slice. `main.yaml` triggers on `pull_request: branches: [main]`, and only PR 1 of this stack
targets `main` — where the workflow is still `main`'s three-job version that has no such job. So the
job never fires anywhere in the stack. Run against the branch tip it reports:

```
Checked: 4  Drifted: 2  Skipped: 3     → exit 1
  data      DRIFT  4 commits, 7 files since 2026-08-10
  database  DRIFT  7 commits, 15 files since 2026-08-10
```

Greening it means updating `.claude/skills/{data,database}/SKILL.md` to reflect schema and data-model
changes that ship in slices 3 and 4 — a content task with its own review needs, deliberately not done
inside a shipping-mechanics change. **It is a trap laid for whoever merges next, so it is named here.**

## Why this PR is red, and why that is expected

Nothing in this slice is executable by CI, and the workflow that would test it targets `main` while
this PR targets its sibling below. The trusted signal for the whole stack is a **full end-to-end suite
run against the development branch tip**, not per-PR CI — per-PR CI on a sibling-based stacked PR can
report green having run nothing.

That run is recorded and it is green: **135 passed, 0 failed, 0 skipped, 0 did-not-run**, with a
database reset from migrations, exactly one fresh dev server, and a preflight that proved the served
application came from the checkout under test.

## Checklist disposition

Every condition of `.agents/code-review-checklist.md` is dispositioned for this slice in
`.planning/phases/151-ship-v0-2-akita-review-stack/151-DISPOSITION.md`. For the record, this slice's
column is **not** the wall of "not applicable" its shape would suggest — **4 MET, 2 DEFERRED, 6 N/A**,
each with a measured reason:

- **Code style — MET by a real gate.** `prettier --check .` genuinely reaches `.claude/`, `.agents/`
  and `CLAUDE.md` (only `.planning/` is ignored, `.prettierignore:36`), and all matched files pass.
  Confirmed by negative control: a deliberately malformed probe file *was* reported, then removed.
- **`any` — DEFERRED, 3 real annotations** in 2 frozen spike sources, none documented. Not edited
  because they are snapshots of experiments that were actually run.
- **Duplication — DEFERRED, 3 genuine groups** out of 16 byte-identical groups; the other 13 are
  determinism evidence where byte-identity *is* the finding, plus `.gitkeep` placeholders.

---

## Addendum — the final re-cut (plan 151-19), and why this note lives here rather than in the repo

**This slice was re-cut one last time at `14afb2d80a9eaf47f8e1e1424ac07f65d1500e2a`**, after plan
151-19 committed its last planning artifact. Criterion 7 is green **at rest** at that commit, not
merely as of a past one:

```
$ bash .planning/phases/151-ship-v0-2-akita-review-stack/scripts/verify-identity.sh \
    feat-gsd-roadmap ship/v0.2-akita-11-planning
changed files: 0
target tree : b606ed169d97276c18da22282cb8ebe18a1887d5
stack  tree : b606ed169d97276c18da22282cb8ebe18a1887d5
BYTE-IDENTICAL                                   # exit 0
```

Catch-all `files=0`. Commit taxonomy over `C1..TIP`: **CONFORMING**, exit 0, 0 shared paths, 0 `[db]`
gaps. All 12 remote refs equal their local tips.

**Why the SHA is written here and not in the repository.** This slice's pathspec is
`.planning .claude .agents CLAUDE.md`, so **the record of the work lives inside the thing it
describes**. Recording this commit's SHA in `151-STACK-MANIFEST.md` would change a file inside the cut
and re-break the identity the line records — the same loop plan 151-17 handed to 151-18 and 151-18
handed to 151-19. It terminates only by moving the record of the last cut *out* of the cut. Recompute
it any time with `git rev-parse ship/v0.2-akita-11-planning`.

**+18 files since the previously published cut**, every one attributed by set difference and nothing
leaving the set: 16 files of the new `.claude/skills/ship-review-stack/` (the ship procedure codified
as a skill, D-25, with all seven scripts carried byte-identically), plus `151-18-SUMMARY.md` and
`151-19-SUMMARY.md`.

**Delta secret rescan, two independent methods, over all 31 files added or modified since the last
scanned cut:** TruffleHog (all detectors, 137 chunks, 1.49 MB) → **0 verified, 0 unverified**. The
independent 24-rule sweep → 34 raw hits, **every one opened and classified, 0 live**: the Supabase
CLI's published loopback default `postgresql://postgres:postgres@127.0.0.1:54322/…` (already shipping
at `apps/supabase/scripts/lint-schema.mjs:27`), pre-existing F-14 prose *describing* a PEM regex, seven
E2E suite-result anchor SHA-256 digests in `STATE.md`, a Svelte prop binding quoted in prose
(`password={passwordOfContext…}`), and long file paths tripping the `long-base64` rule.

### Two things a reviewer should know before merging

- **`skill-drift-check` is now green (F-88 closed).** It exited 1 at the branch tip on `data` and
  `database`, on a job **no PR in this stack shows** — the workflow arrives with slice 10 and fires
  only on PRs targeting `main`. Both skills were resynced against the code they document rather than
  touched to reset the timestamp; in `database`, every one of eleven cited schema filenames had gone
  stale under a renumbering. `bash .claude/scripts/audit-skill-drift.sh` → `Checked: 5  Drifted: 0`.
- **F-89 is open, and it is in PR #868, not this one.** `tests/tests/specs/perf/performance-budget.spec.ts:55`
  reads `## Why a warm-up reload (phase 151, plan 151-18)` — a planning reference that 151-18's own
  fix put into shipped source, three plans after criterion 3 was declared closed. It trips two rows of
  the comment-hygiene gate. It was **not** fixed (slice 05 is published and the force-push
  authorisation for this plan is this branch alone) and **not** absorbed into the gate's expected-red
  state. It needs a decision before or shortly after merge.


.claude/skills/data/object-model.md
```diff
@@ -0,0 +1,151 @@
...
+| ---------------------- | ----------- | ----------------------------------------------- |
+| `Opinion`              | `'opinion'` | Opinion questions normally used for matching    |
+| `Info`                 | `'info'`    | Background/info questions not used for matching |
+| `Default`              | `'default'` | Default type when category type is unspecified  |
+
+Category filtering uses `appliesTo(filters)` on `QuestionAndCategoryBase`, which checks election, constituency, entityType, and electionRound scoping. Questions inherit their category's filters unless `dontInherit: true`.
+
+## Key Relationships
+
+- **Election -> ConstituencyGroup(s) -> Constituency(ies):** Elections have constituency groups; each group contains constituencies the voter chooses from.
```
kaljarv, line 134: Perhaps add in a succinct way:
- The voter can also choose one constituency per election
- In multi-elections settings, the constituency selections cannot be contradictory when multiple elections share the same constituency group or the constituencies are nested

.claude/skills/data/object-model.md
```diff
@@ -0,0 +1,151 @@
...
+- **Constituency -> parentConstituency:** Optional nesting for multi-level elections (e.g., regional + municipal).
```
kaljarv, line 135: Perhaps add: this allows the user to select only the child constiuency and we can imply the parent.

.claude/skills/data/object-model.md
```diff
@@ -0,0 +1,151 @@
...
+- **Nomination links Entity + Election + Constituency:** A Nomination represents an entity nominated in a specific election-constituency pair.
+- **CandidateNomination -> Candidate:** Links to the nominated person.
+- **OrganizationNomination -> Organization:** Links to the party/association; may contain CandidateNominations and FactionNominations as children.
```
kaljarv, line 138: Perhaps add: A very common pattern is a party list of candidates which is an OrganizationNomination with  CandidateNominations as children.

.claude/skills/data/SKILL.md
kaljarv, line 1: Check and possibly refactor all of the existing skills with these recent findings in mind:

Testing Agent-Skills-style progressive disclosure against raw-document navigation and a classical hybrid retriever — across three harnesses, three model families, and InfiniteBench — the authors find disclosure only pays off once the corpus exceeds what the agent can navigate by direct reading. Their sharpest result is about depth: a second routing level never helps and sometimes breaks accuracy outright. One brief description layer pointing at files is the ceiling; nesting indexes under indexes just gives the agent more chances to route wrong before it ever sees content.

For CLAUDE.md / AGENTS.md, the practical translation is to treat the file as a single flat routing layer, not a document tree. Give each area of the repo (or each workflow) a one-to-two-line description plus a direct pointer to the relevant file or directory, and stop there — don't build docs/index.md → docs/backend/index.md → actual content chains, and don't have CLAUDE.md point at intermediate "overview" files whose only job is to point somewhere else. Every intermediate hop is a routing decision the model can fumble, and the paper found those fumbles outweigh any organizational benefit. Inline the small amount of always-relevant content (conventions, build commands) directly, and route to the rest in one hop.

The same rule applies to skills. A SKILL.md's frontmatter description is the routing layer; the body should either contain the instructions or point directly at concrete resource files (references/api.md, scripts/build.py) — not at a second index inside the skill. If a skill has grown a table of contents that links to sub-tables of contents, flatten it: more, smaller skills with sharp one-line descriptions beat one mega-skill with internal hierarchy, because selection happens once at the description layer where the model is most reliable. Spend your effort making those one-liners discriminative (when to use this skill vs. its neighbors) rather than organizing the interior.

The other half of the result is knowing when disclosure is worth anything at all: it "buys context, not intelligence." On a single-document scale, a strong harness that greps and reads on its own gets near-zero benefit from a curated index — so for a small repo, an elaborate CLAUDE.md routing section is mostly wasted effort. The crossover comes when the corpus grows past what direct navigation can handle (multi-repo setups, large internal doc collections): there, raw navigation collapses while one-level disclosure degrades slowly and pulls ahead. A reasonable default: small codebase, keep CLAUDE.md to conventions and commands; large or multi-repo corpus, invest in the flat index — and in both cases, never go deeper than one level. [arXiv:2607.17598](https://arxiv.org/abs/2607.17598)

.claude/skills/database/extension-patterns.md
```diff
@@ -0,0 +1,193 @@
...
+11. **Regenerate database types** `packages/supabase-types/src/database.ts` (repo root relative)
+    - Run: `cd apps/supabase && npx supabase gen types typescript --local > ../../packages/supabase-types/src/database.ts`
+
+12. **Add test data** `tests/database/00-helpers.test.sql`
+    - Add a `test_id` entry for the new entity in the predictable UUID constants section
+    - Add INSERT statement in `create_test_data()` for both Project A and Project B
+    - Project A entities should be published=true; Project B entities should be published=false
+    - Pattern: follow existing entity insertions in create_test_data()
+
+13. **Write pgTAP tests** (see "Adding pgTAP Tests" guide below)
```
kaljarv, line 75: Also, add a step for checking the dev-seed templates whether they need updating as well.

.claude/skills/filters/extension-patterns.md
```diff
@@ -0,0 +1,115 @@
...
+6. **Add tests** `tests/filter.test.ts` (relative to `packages/filters/`)
+   - Import the new filter class from `../src`
+   - Create test entities matching the filter's target type (use `DataRoot` from `@openvaa/data` for question instances)
+   - Test constructor with appropriate options (question or property)
+   - Test `parseValues()` returns correct values from target entities
+   - Test `apply()` with include/exclude rules (or min/max for numeric)
+   - Test `apply()` with `MISSING_FILTER_VALUE` handling (entities without answers)
+   - Test that `filterType` equals the expected `FILTER_TYPE.{Name}Filter` value
+   - Pattern: follow existing `NumberQuestionFilter` or `ChoiceQuestionFilter` test block in `filter.test.ts`
+
```
kaljarv, line 57: Add step: Add e2e tests for the filter, preferably in the full voter journey if applicable.

.claude/skills/matching/SKILL.md
kaljarv, line 1: If it's necessary, add a note to the extension patterns in all skills that the skill itself should be checked afterwards bc they contain listings.

CLAUDE.md
kaljarv, line 1: Evaluate and possibly refactor based on these two findings:

## Evidence check: context files (AGENTS.md/CLAUDE.md) don't improve correctness

A 288-run ablation across Claude Code and Codex on real repositories found no measurable correctness gain from persistent context files.

The study (arXiv, July 28) ran 17 real tasks across 3 repos with gold-standard test grading, comparing runs with and without context files — including each repo's actual human-written AGENTS.md. No configuration converted failures into passes; equivalence testing bounds any effect at ≤10–15 points. The failure analysis is the useful part: agents don't fail from missing repo knowledge (they recover conventions by reading code), they fail on implementation skill — design choices, pattern selection, exact wiring. Practical takeaway: keep context files lean (build/test commands, hard conventions) and stop expecting correctness from prose; invest instead in task decomposition and verification loops (tests, oracles), which target the real failure mode. Code and data are released if you want to reproduce it on your own repos. [Paper](https://arxiv.org/abs/2607.27250)

## Flat hierarchy

Testing Agent-Skills-style progressive disclosure against raw-document navigation and a classical hybrid retriever — across three harnesses, three model families, and InfiniteBench — the authors find disclosure only pays off once the corpus exceeds what the agent can navigate by direct reading. Their sharpest result is about depth: a second routing level never helps and sometimes breaks accuracy outright. One brief description layer pointing at files is the ceiling; nesting indexes under indexes just gives the agent more chances to route wrong before it ever sees content.

For CLAUDE.md / AGENTS.md, the practical translation is to treat the file as a single flat routing layer, not a document tree. Give each area of the repo (or each workflow) a one-to-two-line description plus a direct pointer to the relevant file or directory, and stop there — don't build docs/index.md → docs/backend/index.md → actual content chains, and don't have CLAUDE.md point at intermediate "overview" files whose only job is to point somewhere else. Every intermediate hop is a routing decision the model can fumble, and the paper found those fumbles outweigh any organizational benefit. Inline the small amount of always-relevant content (conventions, build commands) directly, and route to the rest in one hop.

The same rule applies to skills. A SKILL.md's frontmatter description is the routing layer; the body should either contain the instructions or point directly at concrete resource files (references/api.md, scripts/build.py) — not at a second index inside the skill. If a skill has grown a table of contents that links to sub-tables of contents, flatten it: more, smaller skills with sharp one-line descriptions beat one mega-skill with internal hierarchy, because selection happens once at the description layer where the model is most reliable. Spend your effort making those one-liners discriminative (when to use this skill vs. its neighbors) rather than organizing the interior.

The other half of the result is knowing when disclosure is worth anything at all: it "buys context, not intelligence." On a single-document scale, a strong harness that greps and reads on its own gets near-zero benefit from a curated index — so for a small repo, an elaborate CLAUDE.md routing section is mostly wasted effort. The crossover comes when the corpus grows past what direct navigation can handle (multi-repo setups, large internal doc collections): there, raw navigation collapses while one-level disclosure degrades slowly and pulls ahead. A reasonable default: small codebase, keep CLAUDE.md to conventions and commands; large or multi-repo corpus, invest in the flat index — and in both cases, never go deeper than one level. [arXiv:2607.17598](https://arxiv.org/abs/2607.17598)

kaljarv (commented):
Checked.

PR #875, opened 2026-08-18
fix: collected fixes from 2026 deployments

kaljarv:
Contains collected fixes and features from 2026 VAA deployments. See the individual commit messages for more information.


kaljarv:
## Code review

Found 5 issues:

1. `QuestionExtendedInfo` switched to a required `question` prop, but its only consumer `QuestionExtendedInfoDrawer.svelte` was not updated and still passes `title`/`info`/`infoSections`. `question` is therefore `undefined` and the unguarded `const { info } = question;` in `QuestionExtendedInfo.svelte:41` throws `TypeError: Cannot destructure property 'info' of 'undefined'` when a voter opens the question info drawer. `svelte-check` also reports this: `Object literal may only specify known properties, and '"info"' does not exist in type 'QuestionExtendedInfoProps'`.

https://github.com/OpenVAA/voting-advice-application/blob/cc6635a5c86b7f41b64501c85fb834ba249a0b7f/frontend/src/lib/components/questions/QuestionExtendedInfoDrawer.svelte#L37-L45

2. The new Arguments expander is nested inside the `{#if infoSections?.length}` block, so a question that has `customData.arguments` but no `infoSections` never renders its arguments. `arguments` and `infoSections` are independent optional fields in `customData.type.ts` and are produced by two separate admin tools (`condenseArguments.ts` vs `generateQuestionInfo.ts`), so this state is reachable.

https://github.com/OpenVAA/voting-advice-application/blob/cc6635a5c86b7f41b64501c85fb834ba249a0b7f/frontend/src/lib/components/questions/QuestionExtendedInfo.svelte#L49-L77

3. `TITLE_KEYS` maps the pro/con types to the opposite translation keys: `BooleanCons`/`LikertCons` -> `questions.arguments.pro` and `BooleanPros`/`LikertPros` -> `questions.arguments.con`. With `en` (`"con": "Against"`, `"pro": "For"`), `sv`, `da` and `et`, con arguments render under "For" and pro arguments under "Against". `CategoricalPros -> proCategory` is mapped correctly, so this is not a deliberate inversion.

https://github.com/OpenVAA/voting-advice-application/blob/cc6635a5c86b7f41b64501c85fb834ba249a0b7f/frontend/src/lib/components/questions/QuestionArguments.svelte#L36-L44

4. The Finnish `arguments` values are swapped relative to their keys (`"con": "Puolesta"` = for, `"pro": "Vastaan"` = against), unlike the other four locales, and inconsistent with `"proCategory": "Puolesta: {option}"` in the same file. This cancels out issue 3, so Finnish happens to render correctly while `en`/`sv`/`da`/`et` render inverted -- exactly one of issues 3 and 4 should be fixed.

https://github.com/OpenVAA/voting-advice-application/blob/cc6635a5c86b7f41b64501c85fb834ba249a0b7f/frontend/src/lib/i18n/translations/fi/questions.json#L8-L14

5. `sortArguments` sorts in the opposite direction to its own doc comment. `Sort arguments to show cons before pros.` but returning `1` when `a` is a cons argument places cons *after* pros. (The comparator is also non-symmetric: for two cons arguments it returns `1` in either order.)

https://github.com/OpenVAA/voting-advice-application/blob/cc6635a5c86b7f41b64501c85fb834ba249a0b7f/frontend/src/lib/components/questions/QuestionArguments.svelte#L44-L54

🤖 Generated with [Claude Code](https://claude.ai/code)


<sub>- If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

